PR #100, opened 2013-12-24
Minor additions/changes that make makefile more flexible

dastergon:
Hello, I'd like to propose these changes. I think renaming ZSHDIR from vendor-completion into site-functions makes it more generic and software packagers will not need to patch it. Same for the DOCDIR path, adding version next to the vcsh name makes it more flexible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
+self=vcsh
+version=20131214
 PREFIX=/usr
 DOCDIR_PREFIX=$(PREFIX)/share/doc
-DOCDIR=$(DOCDIR_PREFIX)/$(self)
-ZSHDIR=$(PREFIX)/share/zsh/vendor-completions
+DOCDIR=$(DOCDIR_PREFIX)/$(self)-$(version)
+ZSHDIR=$(PREFIX)/share/zsh/site-functions
 RONN ?= ronn
 
-self=vcsh
 manpages=$(self).1
 all=test manpages
 


leycec:
@dastergon This is *mostly* good stuff. Gentoo's [official `vcsh` ebuild](https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/dev-vcs/vcsh/vcsh-1.20150502.ebuild?view=markup), for example, dynamically patches `Makefile` in exactly this way. To no one's surprise, I note that you're a [Gentoo developer](http://blog.dastergon.gr).

Unfortunately, it's also outdated stuff. The merge conflicts are troubling, but it's the `version=20131214` line that concerns me. No one wants to keep two version identifiers in sync – and, let's be honest, **no one will.**

Instead, that variable should simply capture the output of `vcsh version`: e.g.,

    version=`./vcsh version | head -n 1 | cut -d ' ' -f 2`

In any case, thanks for the concise contribution! :hamster:

leycec:
Disappointingly, it would appear that [unversioned documentation directories](https://fedoraproject.org/wiki/Changes/UnversionedDocdirs) are the norm:

> By having the documentation directories unversioned, they and the files in them become inherently easier/stabler to bookmark, cross-reference from other docs, Systemd unit Documentation= etc. Doing so also makes Fedora more consistent with **other major distributions that already do this** (at least Debian, Mandriva, openSUSE, and Ubuntu at time of writing).

Gentoo requires versioned documentation directories, of course. But that sadly appears to be the exception. (For those who know and love the little yellow penguin that could, that's probably no surprise.)

This suggests that this pull request could probably be reduced to a single line correcting the `${ZSHDIR}`. **Brevity.** It rocks.

leycec:
This pull request has been subsumed by #167 and can *probably* be safely closed. Thanks again for the initial inspiration, [dastergon](https://github.com/dastergon).

RichiH:
True; thanks :)

Issue #94, opened 2013-12-09
support for recursive clone?


gliptak:
I'm trying to run

```
 $ vcsh version
 vcsh 1.20130724
 $ vcsh clone --recurse-submodules https://github.com/spf13/spf13-vim.git
```

and receive the following error

```
fatal: Could not switch to 'https://github.com/spf13/': No such file or directory
vcsh: fatal: no repository found for 'spf13-vim'
```

Am I using the right syntax here?

Thanks

gliptak:
With current version:
```
$ vcsh version
vcsh 1.20151229
git version 2.7.4
$ vcsh clone --recurse-submodules https://github.com/spf13/spf13-vim.git
fatal: Could not switch to 'https://github.com/spf13/': No such file or directory
fatal: not in a git directory
fatal: not in a git directory
fatal: not in a git directory
fatal: not in a git directory
vcsh: fatal: no repository found for 'spf13-vim'
```

klaernie:
@gliptak I think leaving the vcsh repo name out and giving options to the `git clone` underneath is mutually exclusive. vcsh sees two arguments passed in, so it treats `--recurse-submodules` as the URL and `https://github.com/spf13/spf13-vim.git` as the repository name.

@RichiH I think `init()` during init it takes the URL as the `$GIT_DIR`:
```
…
+ init
+ hook pre-init
+ [ -x /home/kandre/.config/vcsh/hooks-enabled/pre-init* ]
+ continue
+ [ -x /home/kandre/.config/vcsh/hooks-enabled/spf13-vim.pre-init* ]
+ continue
+ [ ! -e https://github.com/spf13/spf13-vim.git ]
+ mkdir -p /home/kandre
+ cd /home/kandre
+ git init --shared=0600
fatal: Could not switch to 'https://github.com/spf13/': No such file or directory
+ upgrade
+ hook pre-upgrade
```

gliptak:
@klaernie Is there a different syntax to be used? Thanks

klaernie:
@gliptak just `vcsh clone https://github.com/spf13/spf13-vim.git`, then `vcsh spf13-vim submodule update --init --recursive`. This will simply break what you attempted into two steps, as `vcsh clone` is not the same as `git clone`. I tried this with a temporary  `$HOME` and it worked flawlessly.

gliptak:
Thanks @klaernie

PR #136, opened 2014-06-03
Made delete() handle submodules correctly

12qu:
Previously delete() did not use the `-r` flag when invoking rm, which
meant that it could not remove any submodules in a repo (since these
appear as directories in `git ls-files`). This patch adds that flag to
fix this.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -196,7 +196,7 @@ To continue, type 'Yes, do as I say'"
 	read answer
 	[ "x$answer" = 'xYes, do as I say' ] || exit 16
 	for file in $files; do
-		rm -f $file || info "could not delete '$file', continuing with deletion"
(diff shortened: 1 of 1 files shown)


gliptak:
https://github.com/RichiH/vcsh/issues/94

alerque:
@RichiH I think this of the last of the PRs that got closed by GitHub when we renamed the default branch because the downstream fork doesn't exist.

I would resurface it as a new PR, but in this case I am still in agreement with myself from 7 years ago and am not in favor of this being merged at all. As far as I'm concerned I would call this closure good. If anything we could open an *issue* to track the mentioned problem, but I don't think this is an appropriate solution.

RichiH:
Agreed; let's leave it closed. If someone needs submodules deleted, they need to be enumerated and dealt with piece by piece; `-f` is not safe.

PR #145, opened 2015-01-09
Drop dependency on ronn, manpage is truly static

roman-neuhauser:
While the man.7 roff macro set is presentation-level and the source
is ugly, mdoc.7 (if your man page sucks, there's an excellent manual
at [1]) is mostly semantic and not as intrusive.  Readability is half
way between man.7 and markdown.

ronn is a problematic dependency as it's not widely packaged and
requires opening the rubygems can of worms.  This is mitigated by the
manpage-static branch somewhat, OTOH that approach also has its price:
various special cases in packaging code to fix up version strings,
tarball and directory names, etc.), and complicates the release process.

[1] http://mdocml.bsd.lv/man/mdoc.7.html


RichiH:
Thanks. I will need to mull this over.

Maybe I will move the releases to a branch with static manpage and keep the markdown in a pre-release branch. I am not sure yet, but I think I want to keep markdown internally.

RichiH:
I think I will branch `release` off of `master`, merge `manpage-static` and `release`, and then release in `release`. That way, I can drop the external dependency while still maintaining my workflow with Markdown.

vdemeester:
This seems ok. You'll need ronn if you want to install vcsh from git (in unstable-mode) and no dependency to ronn for the release. I like it :P

RichiH:
https://github.com/RichiH/vcsh/pull/145#issuecomment-70009500 has been implemented in 24949d03789aab3a6a4a5ba8c2964bf4415f9832 on May 2nd.

PR #148, opened 2015-02-27
vcsh shortstat

dottedmag:
short status only shows repositories with changes, which makes it
suitable to be run during login shell startup.

Regular status:
    % ./vcsh status
    basic:
    Ahead of origin/master by        1 commits
    
    emacs:
    
    keys:
     M ../../.gnupg/gpg.conf
     M ../../.ssh/known_hosts
    
    osx:
    
    private:

Short status:

    % ./vcsh shortstat
    basic:
    Ahead of origin/master by        1 commits
    keys:
     M ../../.gnupg/gpg.conf
     M ../../.ssh/known_hosts

### DIFF
--- a/t/300-add.t
+++ b/t/300-add.t
@@ -29,5 +29,11 @@ A  a
 
 ", 'Adding a file works';
 
+my $output = `.././vcsh status --terse`;
+
+ok $output eq "test1:
+A  a
+", 'Terse output works';
+
 done_testing;
 
--- a/vcsh
+++ b/vcsh
@@ -119,7 +119,8 @@ help() {
           <newname>     Rename repository
    run <repo> \\
        <command>        Use this repository
-   status [<repo>]      Show statuses of all/one vcsh repositories
+   status \\
+     [--terse] [<repo>] Show statuses of all/one vcsh repositories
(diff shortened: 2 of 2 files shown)


dottedmag:
Updated to spell `stat --terse` instead of `shortstat`

RichiH:
As per IRC: Thanks for the patch, I think it's very useful. Maybe I will even make that the default.

That being said, your latest PR makes the tests fail. Can you fix that, please? Ideally, you could also provide a short test for the new output.

Thanks!

RichiH:
https://travis-ci.org/RichiH/vcsh/builds/52450207

dottedmag:
I see, it is due to that:

    +			STATUS=$(status_helper $VCSH_REPO_NAME "-c color.status=always")

I store output of `git status` into a variable and in order to get a coloring I force it unconditionally. Tests fail because they don't expect the output to be colored.

Proper way would be to detect whether current stdout is tty and enable coloring in this case. How one does that properly in shell?

dottedmag:
How one does that properly in shell?
    test -f <fd>

Updated test pushed.

RichiH:
Thanks!

And sorry for taking so long.. Real life gets more and more crazy...

PR #149, opened 2015-03-02
Fix tests under git-2.2

dottedmag:
I have noticed that one of the tests failed under git-2.2

### DIFF
--- a/t/100-init.t
+++ b/t/100-init.t
@@ -25,7 +25,6 @@ ok $output eq "test1:\n\n", 'Our new repo is there';
 chdir $ENV{"HOME"} . '/.config/vcsh/repo.d/test1.git/' or die $!;
 
 ok -f 'HEAD';
-ok -d 'branches';
 ok -f 'config';
 ok -f 'description';
 ok -d 'hooks';


RichiH:
Could you try to make this dependent on version? I.e. with legacy Git, test for it, with 2.2 you don't?

dottedmag:
It's been deprecated since 2008, maybe it's easier to just drop it? https://github.com/git/git/commit/1dd1239aa33ddd7f159cd183338ef6f71298e29a

jwhitley:
+1 on accepting this PR as-is.

I'm on git 2.4.5, and just got hit by this while trying to install and test the `feature/foreach` branch.

vcsh (rightly) doesn't directly depend on the actual presence or absence of this directory, only on git's external behavior.  The whole `t/100-init.t` suite only has meaning as a collective because each test is fingerprinting the implementation details of git's repo structure.  Moreover, I can't imagine a case where `vcsh init` causes an (old, < v2.2) version of git to pass this suite *except* for `ok -d 'branches';`.

RichiH:
Thanks :)

PR #156, opened 2015-04-14
vcsh: list_untracked(): remove --nocheck-order

elyscape:
`--nocheck-order` is only supported by GNU `comm`, but it isn't
necessary anyway because we sort the inputs first.


elyscape:
Sorry, wrong button.

RichiH:
Thanks.

Issue #54, opened 2013-07-01
vcsh rename and core.excludesfile


a-sk:
vcsh rename should also rename core.excludesfile repo's config section

PR #162, opened 2015-05-19
keep gitignore & gitattributes on upgrade/rename

### LINKED ISSUE #37: rename command should take care of .gitignore file
I just noticed that 'vcsh rename oldname newname' neither renames the gitignore file in ~/.gitignore.d, nor updates the core.excludesfile directive in the renamed repositories' config.

As far as I understand,  this leads to some unexpected behavior when writing a new gitignore after renaming or adding another repository named 'oldname' after issuing the above command, right?

guyhughes:
fixes #37, further to #54 -- I feel that the user expects to keep the .gitignore.d/foo and .gitattributes.d/foo files in effect when renaming repo foo to bar.  Currently `upgrade()` abandons the foo files and updates .git to point to bar files.

I've implemented renaming the files with [`mv`](http://pubs.opengroup.org/onlinepubs/9699919799//utilities/mv.html) in `upgrade()` so that even if `core.excludesfile` or `core.attributesfile` points elsewhere, they can be moved into the recommended location. 

`mv` will exit with an error instead of overwriting any file, and I've handled showing that to the user. 

I've only played with this a bit. Open for improvements.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -410,10 +410,34 @@ upgrade() {
 	elif [ ! "x$VCSH_WORKTREE" = 'xrelative' ]; then
 		git config core.worktree "$VCSH_BASE"
 	fi
-	[ ! "x$VCSH_GITIGNORE" = 'xnone' ] && git config core.excludesfile ".gitignore.d/$VCSH_REPO_NAME"
-	[ ! "x$VCSH_GITATTRIBUTES" = 'xnone' ] && git config core.attributesfile ".gitattributes.d/$VCSH_REPO_NAME"
-	git config vcsh.vcsh 'true'
 	use
+	if [ ! "x$VCSH_GITIGNORE" = 'xnone' ]; then
+    old_gitignore_file=$(git config core.excludesfile)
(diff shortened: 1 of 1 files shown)


alerque:
Hey thanks for taking the time to contribute this. I did some cleanup work to the coding style and tweaked some shell scripting issues. I haven't had a chance to put it through any paces or really figure out if the approach is right, but I can't push to this PR (it was opened under GitHub's legacy system, you *might* have a check box to enable maintainers to modify the PR) but wanted to note that I stashed the WIP on this in [this branch](https://github.com/RichiH/vcsh/compare/main...alerque:feature/keep-gitignore).

alerque:
See also extra discussion in #163 and particular the alternative implementation in [this branch](https://github.com/RichiH/vcsh/tree/fix/rename_gitignore_gitattributes). That should be compared as part of reviewing this.

alerque:
Hey @guyhughes just a gentle ping if you're around and can check for that "allow maintainers..." checkbox in the sidebar that would facilitate moving this along finally.

PR #165, opened 2015-05-25
t/100-init.t: Fix failure due to locale

vlow:
When running the t/100-init.t test, it relies on an exact response from git. This fails, if the locale is different from the English default. This change fixes the bug by enforcing the English default locale for this test.


RichiH:
Thanks!

Issue #152, opened 2015-04-10
Unable to install vcsh after downloading the git repo.


absorber:
I followed the prerequisites stated [here](https://github.com/RichiH/vcsh/blob/master/doc/INSTALL.md) and even tried this after installing `ruby-ronn`.

This is what I did: 

```
$ git clone https://github.com/RichiH/vcsh
Cloning into 'vcsh'...
remote: Counting objects: 2551, done.
remote: Total 2551 (delta 0), reused 0 (delta 0), pack-reused 2551
Receiving objects: 100% (2551/2551), 586.67 KiB | 514.00 KiB/s, done.
Resolving deltas: 100% (1377/1377), done.
Checking connectivity... done.
```

then, `cd vcsh/`

```
$ sudo make install
[sudo] password for me: 
t/000-tear-env.t ... Can't locate Shell/Command.pm in @INC (@INC contains: /etc/perl /usr/local/lib/perl/5.14.2 /usr/local/share/perl/5.14.2 /usr/lib/perl5 /usr/share/perl5 /usr/lib/perl/5.14 /usr/share/perl/5.14 /usr/local/lib/site_perl .) at t/000-tear-env.t line 6.
BEGIN failed--compilation aborted at t/000-tear-env.t line 6.
t/000-tear-env.t ... Dubious, test returned 2 (wstat 512, 0x200)
No subtests run 
t/001-setup-env.t .. Can't locate Test/Most.pm in @INC (@INC contains: /etc/perl /usr/local/lib/perl/5.14.2 /usr/local/share/perl/5.14.2 /usr/lib/perl5 /usr/share/perl5 /usr/lib/perl/5.14 /usr/share/perl/5.14 /usr/local/lib/site_perl .) at t/001-setup-env.t line 6.
BEGIN failed--compilation aborted at t/001-setup-env.t line 6.
t/001-setup-env.t .. Dubious, test returned 2 (wstat 512, 0x200)
No subtests run 
t/100-init.t ....... Can't locate Test/Most.pm in @INC (@INC contains: /etc/perl /usr/local/lib/perl/5.14.2 /usr/local/share/perl/5.14.2 /usr/lib/perl5 /usr/share/perl5 /usr/lib/perl/5.14 /usr/share/perl/5.14 /usr/local/lib/site_perl .) at t/100-init.t line 7.
BEGIN failed--compilation aborted at t/100-init.t line 7.
t/100-init.t ....... Dubious, test returned 2 (wstat 512, 0x200)
[...]

Issue #41, opened 2013-04-20
TEST: Create test that tagged version is returned by `vcsh version`


RichiH:
Else, I may/will forget to update this at some point in time.

alerque:
Yup. Like with the current safe-harbor version tag ;-)

So....as written this is pretty hard to test for because by the time a test (at least remotely in CI) would fail the tag has already been created. A local check that verified this after tagging would do something. If you ran it. But I digress.

The real question I want to ask is how would you feel if I restructured this a bit to treat it as a *real* software project with a configure and build system instead of a "take it or leave it" script? My personal preference would be to just use autoconf / automake. I realize it's a bit long in the tooth, but it already handles a ton of things that even this project could benefit from. Off the top of my head a couple that come to mind:

* Move versioning to the configure system, optionally make it auto-generated based on the tag situation
* Local path discovery to things that may or may not be we have hard coded guesses.
* Man pages would be generated at tag time and automatically included in the source tarball on releases (made with `make dist`, automatically built and attached to tags on GitHub), `ronn` would only be required to build from Git, not from source packages.
* Build time dependency checks are better than run time failures.
* Easier on distro packagers. Right now packaging on Arch requires patching or goofy command invocations that you can only discover by parsing the Makefile. Configure flags are much more discoverable.

I realize there is an existing PR already dealing with many of the current Makefile ailments (most of which I agree with, yes I see you #167) but none of them go so far as actually introducing a configure step or making `vcsh` a generated script or doing any proper system discovery stuff.

Are there any systems currently supported by VCSH that have a POSIX shell plus `make` that **don't** have autoconf/automake? Would you consider distributing source packaging instead of flat `git archive` tarballs a bad thing (automated of course)?

Thoughts @RichiH?

PR #167, opened 2015-06-08
Makefile Portability and Safety Improvements

leycec:
This pull request addresses a depressing cornucopia of pressing Makefile issues, fixing #152 (and probably more) and subsuming pull requests #100 and arguably #145 (and probably more).

For the good of all that is portable and safe, let's get to it.
## A Few Problems

The laundry list of dirty ills includes:
- Installation requires unit tests to be run, thus requiring a [full Perl installation](#152) complete with the `Test::Most` and `Shell::Command` modules **despite `vcsh` itself not requiring Perl**.
- Installation requires man pages to be generated, thus requiring [`ronn` and hence Ruby](#145) **despite `vcsh` itself not requiring Ruby**.
- Zsh completions are installed to the [Debian-specific directory](http://www.zsh.org/mla/workers/2013/msg00622.html) `$(PREFIX)/share/zsh/vendor-completions` **regardless of whether Zsh is even installed**.
- Standard directories (e.g., `$(PREFIX)/bin`) are unsafely created when not found. If `/usr/bin` no longer exists, **something is horribly wrong**. Silently creating that directory (probably without the correct ownership or permissions) and blithely stumbling on as if everything is [hunky dory](https://www.youtube.com/watch?v=YQTENuQYgjM) is _not_ the sane response.
- Standard directories are hardcoded rather than exposed to the user as overridable variables (e.g., `$(PREFIX)/bin` rather than `$(BINDIR)`), contravening [GNU makefile conventions](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) and complicating the quiet, desperate lives of downstream packagers.
- Abstract makefile targets (i.e., targets with no existing path of the same basename) are _not_ listed in the `.PHONY` pseudo-target. While this isn't currently a concern, it is a plausible [ticking time bomb](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) should any path with the same basename as an existing target be added to the `vcsh` codebase. 
## A Few Solutions

Our laundry list of sanitary fixes includes:
- **Divorcing installation from unit test running and man page generation.** Unit tests remain runnable via `make test` and man pages generable via `make manpages`, but neither are required to install `vcsh` anymore. For sanity, any man pages that _have_ been generated will still be installed when `sudo make install` is run. That's nice.
- **Zsh completions are installed to the [platform-agnostic directory](http://zsh.sourceforge.net/releases.html)** `$(PREFIX)/share/zsh/site-functions` _only_ if that directory and hence Zsh is installed.
- **Installation halts with a fatal error if any standard directory does not already exist.** Specifically, all such directories are explicitly suffixed by `/` when passed to `install`.
- **Standard directories are exposed to the user as overridable variables** with conventional names (e.g., `$(BINDIR)`, `$(MANDIR)`). 
- **Abstract makefile targets are listed in the `.PHONY` pseudo-target.**

Also, all targets now have human-readable comments.

The gnomes have learned a new way to say, ["Hooray!"](https://www.youtube.com/watch?v=em4Yl8qIm9Q) :ghost:

### DIFF
--- a/BSDmakefile
+++ b/BSDmakefile
@@ -1,0 +1,17 @@
+# BSD make-specific makefile, read only by BSD make and hence containing
+# BSD make-specific extensions (e.g., "ifeq (...)", "$(shell ...)").
+
+#FIXME: Conditionalize according to the "GNUmakefile" approach.
+
+# Default target, running all preparatory installation targets without actually
+# installing. This *MUST* be the first non-"."-prefixed target in this makefile.
+all: manpages test
+
+# Install vcsh and supplementary artifacts.
+install: all install-common
+# Install man pages and zsh completions.
+	install -m 0644 $(manpages) $(DESTDIR)$(MANDIR)/man1/
+	install -m 0644 $(zshcompletions) $(DESTDIR)$(ZSH_COMPLETIONS_DIR)/
+
+# Define all platform-agnostic variables and rules.
+.include "Makefile"
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,0 +1,71 @@
+# GNU make-specific makefile, read only by GNU make and hence containing
+# GNU make-specific extensions (e.g., "ifeq (...)", "$(shell ...)").
+
+# Define all platform-agnostic variables and rules.
+include Makefile
+
+# Absolute path of zsh's completions directory. Since Debian-based Linux distros
+# prefer a different directory to that of most platforms, test for the existence
+# of such directory before falling back to a platform-agnostic directory.
+ifneq ($(wildcard $(DESTDIR)$(ZSHDIR_PREFIX)/vendor-completions),)
+ZSH_COMPLETIONS_DIR:=$(ZSHDIR_PREFIX)/vendor-completions
+else ifneq ($(wildcard $(DESTDIR)$(ZSHDIR_PREFIX)/site-functions),)
+ZSH_COMPLETIONS_DIR:=$(ZSHDIR_PREFIX)/site-functions
+endif
+
+# List of targets to be run both when no target is specified and when the
+# "install" rule is run.
+all_targets:=
+
+# Non-empty only if the "manpages" target is listed in $(all).
+is_manpaging=
+
+# Non-empty only if the "test" target is listed in $(all).
+is_testing=
+
+# If Ronn is in the current $PATH, generate Ronn-based man pages by default.
+ifneq ($(shell command -v "$(RONN)" 2>/dev/null),)
+all_targets+= manpages
+is_manpaging:=true
+endif
+
+# If Git, Perl, and the Perl-based unit test runner "prove" are all in the
+# current $PATH and the requisite Perl modules are in the @INC module path, run
+# Perl-based unit tests by default.
+ifneq ($(shell command -v git 2>/dev/null),)
+ifneq ($(shell command -v perl 2>/dev/null),)
+ifneq ($(shell command -v prove 2>/dev/null),)
+ ifeq ($(shell perl -e 'use Test::Most; use Shell::Command;' 2>&1),)
+all_targets+= test
+is_testing:=true
+endif
+endif
+endif
+endif
+
+# Default target, running all preparatory installation targets without actually
+# installing. This *MUST* be the first non-"."-prefixed target in this makefile.
+all: $(all_targets)
+# Notify the user of skipped targets.
+ifeq ($(is_manpaging),)
+	@echo 'Skipping man page generation: "$(RONN)" not found.' 1>&2
+endif
+ifeq ($(is_testing),)
+	@echo 'Skipping unit tests: "git", "prove", and/or Perl modules "Test::Most" and "Shell::Command" not found.' 1>&2
+endif
+
+# Coerce the default target to be the prior target (rather than the first rule
+# defined by the makefile included above).
+.DEFAULT_GOAL:=all
+
+# Install vcsh and supplementary artifacts.
+install: all install-common
+# If the man page has been generated, install it during installation.
+ifneq ($(wildcard $(manpages)),)
+	install -m 0644 $(manpages) $(DESTDIR)$(MANDIR)/man1/
+endif
+
+# If zsh's completions directory exists, install ours there during installation.
+ifneq ($(ZSH_COMPLETIONS_DIR),)
+	install -m 0644 $(zshcompletions) $(DESTDIR)$(ZSH_COMPLETIONS_DIR)/
+endif
--- a/Makefile
+++ b/Makefile
@@ -1,48 +1,75 @@
-PREFIX?=/usr
-DOCDIR_PREFIX=$(PREFIX)/share/doc
-DOCDIR=$(DOCDIR_PREFIX)/$(self)
-ZSHDIR=$(PREFIX)/share/zsh/vendor-completions
-RONN ?= ronn
+# make-agnostic makefile, read by all make flavours including GNU make and hence
+# consisting of only POSIX make-compliant syntax.
+#
+# This makefile defines all variables and rules except those required by and
+# including the "all" and "install" rules, whose complexity necessitates
+# conditional logic inexpressible in POSIX make-compliant syntax.
 
 self=vcsh
+
+PREFIX?=/usr
+BINDIR=$(PREFIX)/bin
+DOCDIR_PREFIX=$(PREFIX)/share/doc
+DOCDIR=$(DOCDIR_PREFIX)/$(self)
+MANDIR=$(PREFIX)/share/man
+RONN ?= ronn
+
+# Basename of the man page to be generated.
 manpages=$(self).1
-all=test manpages
 
-all: $(all)
+# Basename of our zsh completions script.
+zshcompletions=_$(self)
 
-install: all
-	install -d $(DESTDIR)$(PREFIX)/bin
-	install -m 0755 $(self) $(DESTDIR)$(PREFIX)/bin
-	install -d $(DESTDIR)$(PREFIX)/share/man/man1
-	install -m 0644 $(manpages) $(DESTDIR)$(PREFIX)/share/man/man1
+# Absolute path of zsh's top-level data directory.
+ZSHDIR_PREFIX=$(PREFIX)/share/zsh
+
+# Absolute path of zsh's completions directory.
+ZSH_COMPLETIONS_DIR=$(ZSHDIR_PREFIX)/site-functions
+
+# List of abstract target names not corresponding to actual paths, preventing
+# conflicts with any such paths and improving efficiency in general.
+.PHONY: all clean install install-common manpages moo purge test uninstall
+
+# Install all paths *NOT* requiring conditional and hence non-POSIX logic.
+install-common:
+# Install the "vcsh" command.
+	install -m 0755 $(self) $(DESTDIR)$(BINDIR)/
+
+# Install documentation.
 	install -d $(DESTDIR)$(DOCDIR)
-	install -m 0644 README.md $(DESTDIR)$(DOCDIR)
-	install -m 0644 doc/hooks $(DESTDIR)$(DOCDIR)
-	install -d $(DESTDIR)$(ZSHDIR)
-	install -m 0644 _$(self) $(DESTDIR)$(ZSHDIR)
+	install -m 0644 README.md $(DESTDIR)$(DOCDIR)/
+	install -m 0644 doc/hooks $(DESTDIR)$(DOCDIR)/
 
+# Generate the manpage.
 manpages: $(manpages)
 
 $(self).1: doc/$(self).1.ronn
 	$(RONN) < doc/$(self).1.ronn > $(self).1 || rm $(self).1
 
+# Remove the generated manpage.
 clean:
 	rm -rf $(self).1
 
+# Remove all previously installed paths.
 uninstall:
-	rm -rf $(DESTDIR)$(PREFIX)/bin/$(self)
-	rm -rf $(DESTDIR)$(PREFIX)/share/man/man1/$(self).1
+	rm -rf $(DESTDIR)$(BINDIR)/$(self)
+	rm -rf $(DESTDIR)$(MANDIR)/man1/$(self).1
 	rm -rf $(DESTDIR)$(DOCDIR)
-	rm -rf $(DESTDIR)$(ZSHDIR)/_$(self)
+	rm -rf $(DESTDIR)$(ZSH_COMPLETIONS_DIR)/$(zshcompletions)
 
-# Potentially harmful, used a non-standard option on purpose.
-# If PREFIX=/usr/local and that's empty...
+# FIXME: The "--ignore-fail-on-non-empty" option is GNU-specific.
+
+# Remove all previously installed paths and all empty parent directories of
+# these paths. Since this is potentially harmful, this target has intentionally
+# been given a non-standard name. For example, if $(PREFIX) is "/usr/local" and
+# now empty due to uninstalling vcsh, this target would remove "/usr/local"!
 purge: uninstall
-	rmdir -p --ignore-fail-on-non-empty $(DESTDIR)$(PREFIX)/bin/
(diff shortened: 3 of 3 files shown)


RichiH:
Thanks for the PR!

> Installation requires unit tests to be run

I agree that this could be done away with in the `release` branch. If you are using `master`, you are expected to know how to fix this yourself.
Ideally, try to run the tests and just carry on if possible as per https://github.com/RichiH/vcsh/issues/152#issuecomment-92216387

> Installation requires man pages to be generated

`release` branch removed that some time ago. `master` still has it, but if you use that, you should be able to cope.

> Zsh completions are installed to the Debian-specific directory 

True, but that needs to check for both Debian and non-Debian directories.

> Standard directories (e.g., $(PREFIX)/bin) are unsafely created

Valid point.

> Standard directories are hardcoded rather than exposed to the user as overridable variables

Valid point.

> Abstract makefile targets

Neat. I didn't know that.

> Also, all targets now have human-readable comments.

I thought there was a law or three against easily parsed Makefiles ;)


Can you update the PR or create a new one? Thanks :)

leycec:
> Thanks for the PR!

Thanks for the one true dotfile manager.

> I agree that this could be done away with in the release branch.

**Ugh.** I'd forgotten that the `release` rather than default `master` branch is the intended branch for end-user installation. Life, why are you not simpler?

>  If you are using master, you are expected to know how to fix this yourself.

Since `master` is (understandably) the default branch, it seems likely that more than a few users will simply clone `master`, run `make install`, and promptly fail. Like, say... **me.** :hankey: 

While this *is* an egregious example of user error, the current repository and makefile design do little to discourage this sort of understandable confusion.

To be sure, I maintain a happy-go-lucky [third-party Gentoo overlay](https://github.com/leycec/raiagent) comprising over 40 open-source packages – most still requiring autotools- and hence makefile-based compilation and installation. None require unit tests to be run prior to installation, even those cloning the HEAD from live Github repositories. I'm reasonably confident that that's *not* standard practice in either open- or closed-source development. Which isn't to say that it's wrong, per say. It might very well be zealously rationale in `vcsh`'s case. But it does contravene basic expectations *and* significantly complicate installation from the default branch.

**This generally makes sense.** Unit tests are themselves prone to spurious failure irrespective of the working state of the codebase they test. Unit tests typically require additional dependencies *not* required by that codebase. Unit tests consume excess time and space. Unit tests are *not* hard compile- or run-time dependencies. In short, unit tests are entirely optional and best left to the user's discretion. (Usually, anyway.)

Some combination of server-side continuous integration (CI) and/or git `update` hooks would seem to be the sounder approach to quality assurance. But `vcsh` already *has* Travis-based CI, and probably hooks as well. (Which is awesome, by the way. Definite kudos on that.)

I'm not trying to grill you, Rich. I abjectly apologize if anything above comes off as unfairly critical. You probably have very sound reasoning for requiring unit tests to be run prior to installation. I know you do! In my case, however, the simplest solution "...to fix this [failing unit tests] yourself" was simply to remove the `all` dependency from the `install` makefile target – which is to say, to install without running unit tests at all.

That said, if you *really*, ***really*** want to enforce unit test execution on `vcsh` installation, I'll definitely revert that back and rebase this. That's no problem. It's just an... *interesting* design choice. I'd be curious to hear your rationale.

leycec:
**Updated.** zsh completions are now [preferentially installed](https://github.com/leycec/vcsh/commit/dc59c12d7a99a23686fdb95d28d4cc7fb660d781#diff-b67911656ef5d18c4ae36cb6741b7965L11) to a Debian-specific directory before falling back to a platform-agnostic directory. God’s in His heaven— All’s right with the world.

RichiH:
> tests

All valid points. And docs or no docs, people will always try to work from `master`, _especially_ the ones who can't cope with perl deps missing.
How about we make tests optional, but carp when they are missing about it would be better to run them?

> completions

Nitpick, but `ZSHDIR` ist sub-optimal, no? `ZSH_COMPLETIONS_DIR` (and `_PREFIX`) are ungainly, but factually correct. If we ever needed to access a different part of the system, which is very unlikely, that variable would already be taken.

leycec:
> Nitpick, but ZSHDIR ist sub-optimal, no? ZSH_COMPLETIONS_DIR (and _PREFIX) are ungainly, but factually correct. If we ever needed to access a different part of the system, which is very unlikely, that variable would already be taken. And yes, I realize that ZSHDIR was my invention, but when we are spring cleaning, anyway...

**Spot on.** `ZSHDIR` *is* inappropriate here – an observation that crossed my mind as well.

When submitting my first pull request to a new project, I'm apprehensive of going... a *little* too far. (Insert flashback of unaccepted pull requests, complete with awkward voice-over.) I dialled it back on this one and retained the existing nomenclature. As you suggest, `ZSH_COMPLETIONS_DIR` is probably the sane choice.

Spring cleaning is in session.

leycec:
**Updated.** `ZSHDIR` is now `ZSH_COMPLETIONS_DIR` and installation dependencies are now conditionally run *only* when such dependencies are likely to succeed, as tiresomely detailed above. On my cobbled-together system, for example, `sudo make install` now yields the following output:

```
"git", "prove", and/or Perl modules "Test::Most" and "Shell::Command" not found; skipping unit tests.
"ronn" not found; skipping man page generation.
install -m 0755 vcsh /usr/bin/
install -d /usr/share/doc/vcsh
install -m 0644 README.md /usr/share/doc/vcsh/
install -m 0644 doc/hooks /usr/share/doc/vcsh/
install -m 0644 _vcsh /usr/share/zsh/site-functions/
```

RichiH:
Thanks; I will prolly be able to review in depth tonight.

As a quick reply: Are you saying that the current Makefile is not BSD-compatible, that your changes are not, or that you are unsure?
`vcsh` is written in POSIX shell on purpose; I would hate to have the Makefile not be portable if it can be avoided at all.

alerque:
Suggestion: use a more standard branch layout. With git branch names (and even the default) are all a matter of convention, but you are both right that the first thing anybody will try to do is clone master and try to run with it. I would rotate the release branch to be master and use a develop branch for ongoing changes in-between releases—_al la_ [git-flow style branch usage](http://nvie.com/posts/a-successful-git-branching-model/).

leycec:
> Thanks; I will prolly be able to review in depth tonight.

No worries. Time is plentiful; life is rich.

> As a quick reply: Are you saying that the current Makefile is not BSD-compatible, that your changes are not, or that you are unsure?

I should have been less blatantly ambiguous.

Let's go with **(B)**, please. The current Makefile appears to be simplistic enough as to be BSD-compatible; my changes, lamentably, are not. This is fixable in one of several ways:

* **Abandoning `make` for an alternative build system.** As a package maintainer, I'm indecently fond of [Cmake](http://www.cmake.org). It makes (...get it, *makes*?) life tractable. It's sensible; it's portable; there are no hidden gotchas. `make` is none of these things.
* **Augmenting `make` with `automake`.** Allow me to sever my femoral artery in this dank corner. Autotools is clearly the [work of the devil](http://voices.canonical.com/jussi.pakkanen/2011/09/13/autotools). I advise having nothing to do with it.
* **Refactoring our single `Makefile` into a GNU-specific `GNUmakefile` and BSD-specific `BSDmakefile`.** GNU `make` searches for and runs `GNUmakefile` *before* falling back to `Makefile`; likewise, BSD `make` searches for and runs `BSDmakefile` *before* falling back to `Makefile`. To preserve sanity, both `GNUmakefile` and `BSDmakefile` are typically designed to include a common `Makefile.common` file containing *only* simplistic statements guaranteed to execute portably.

Short of switching full bore to Cmake (...which wouldn't be the worst long-term move, really), refactoring the current makefile into the three makefiles detailed above is probably our best short-term move. I'd be happy to poke my finger into that disagreeable hornet's nest at some point, although I'm unsure I have the courage, mental fortitude, or available time for that... at the moment, anyway.

> vcsh is written in POSIX shell on purpose; I would hate to have the Makefile not be portable if it can be avoided at all.

I love All Things POSIX (tm), but it pains me to admit that the POSIX committee *really* screwed up `make` standardization. It's well-known to be infeasible to write cross-platform single-file Makefiles for anything but the simplest of toy applications, rendering POSIX standards irrelevant insofar as software deployment is concerned. **Which totally sucks.**

The adjective "anemic" usually comes up in these sorts of vitriolic discussions. To quote the [GNU `make` mailing list](https://lists.gnu.org/archive/html/help-make/2002-10/msg00029.html) from the eldritch days of 2002:

>   g> I am trying to write makefiles that are least dependent on a
>   g> particular make.  The Gnu make manual says:
> 
> Let me prefix this message by saying that what you want to do is going
> to be very frustrating, and you should only proceed if you have _VERY_
> strong reasons you _MUST_ do this, such as external requirements.  The
> POSIX definition of make is very anemic: it's so undemanding that it
> doesn't even provide for an "include" feature, much less many other more
> powerful capabilities.
> 
> It's far, far better for your sanity to forego the idea of writing
> portable makefiles, and use a portable _make program_ instead.  GNU make
> is available in source form and compiles on virtually every version of
> UNIX, plus VMS, DOS, Windows, and Amiga.  Coming shortly (currently
> available as patches) is a port to OS/2.
> 
> Failing that, you'll probably want to use automake or some other
> makefile generating tool: writing POSIX compliant makefiles is tedious
> enough that you won't want to do it by hand.

leycec:
**O.K.** I *really* shouldn't, but I'm going to go ahead and split the GNU-specific portion of my changeset into a new `GNUMakefile`. The existing `Makefile` will remain POSIX- and hence BSD-compatible, which should satisfy everyone and their fetid grandmother.

Due to time constraints (...that I've already blown *way* past), this constitutes my final large-scale update to this pull request. While I'd be happy to continue making minor revisions, I'm afraid that major revisions are beyond the scope of my life. Such is free-market volunteerism. (You get what you pay for.)

RichiH:
Hi,

As i don't know what will go where: Can I still keep the test and manpage
behaviour the same in both Makefiles in one branch if you do that?
As there are a lot of OS X users I need that.

Thanks once again for your efforts,
Richard

Sent by mobile; excuse my brevity.

leycec:
**It is done.** This is probably the end of the open-source road for me and this pull request. In other words, "I drink your milkshake."

leycec:
> As i don't know what will go where: Can I still keep the test and manpage behaviour the same in both Makefiles in one branch if you do that? As there are a lot of OS X users I need that.

Excellent question. Ideally, yes. At the moment, I'm unsure how to reliably write the corresponding logic in BSD `make`-compatible syntax. That's *not* necessarily a bad thing. Here's how this all plays out.

The existing `Makefile` has been split into a new `GNUmakefile` and a new `BSDmakefile`, as discussed. The former has been "conditionalized" to run unit tests and generate and install man pages in an intelligent manner; the latter retains the prior unconditional logic and hence *always* runs unit tests and generates and installs man pages.

OS X users should, therefore, see no change in installation behaviour – unless they manually install GNU `make` (e.g., via `brew`, which isn't terribly arduous). Linux users will, on the other hand, enjoy the fruits of our sweaty labour. In either case, `vcsh` still installs as expected. It just installs a bit more *cleverly* under GNU `make`.

leycec:
**Necro-post.** :scream: 

I now use this branch _exclusively_ under a variety of platforms: Gentoo, Ubuntu 15.10, Windows XP Pro x64, OS X Snow Leopard, and the list just goes on and on. No glaring issues to report yet. All patched makefiles operate as expected – which is really more than I expected.

That said, I have _not_ tested these changes under a BSD derivative. I'm afraid I haven't set up a guest image for that. Since it's unclear when I'll find the time for that, it'd be great if some intrepid community member could test this on everyone's behalf. (You will be much loved.)

RichiH:
I will try to find someone with a BSD willing to test.

alerque:
Hey there! And thanks for this PR. You've covered a lot of the things I've hated about the current makefile for years. I wish I'd actually find this earlier and started using this makefile for distro packaging, even without it being merged!

Caveat: I've only skimmed the discussion above so forgive me if I missed something.

I'd like to propose taking this a step further. Most of the changes in this Makefile are good, but I'd like to move a *couple* of the logic bits to a configure step using autoconf/automake. See [my comment in #41](https://github.com/RichiH/vcsh/issues/41#issuecomment-811079857) for a bit more background on what I have in mind.

Is this a stumbling block for anything I don't know about? Obviously the trade off would be that you would need to use a source tarball generated by `make dist` instead of some release branch to install without extra dependencies and Automake/Autoconf would be a new dependency for a) developers and b) anyone wanting to build from a random Git HEAD commit instead of a release tag. Is this direction objectionable if I were to code it up (on top of these other Makefile fixes of course).

RichiH:
I really dislike autoconf / automake; `meson` seems to be the New Cool Thing. But: I can see why it makes sense.

I think merging this one into mainline so people who want it can use it and then building on top makes sense. Thoughts?

alerque:
> I really dislike autoconf / automake; `meson` seems to be the New Cool Thing. But: I can see why it makes sense.

I kind of suspected that might be the case. I've only recently come to a point where I don't despise it, and that after 25+ years of using it!

Yes `meson` is one of the current hotnesses, but as far as I'm concerned it's a non-starter for this project because it's an extra dependency for any platform that doesn't have prebuilt packages. The advantage of Autotools is that only the developers need it at all. Everybody else only needs make + coreutils because it ships a *very* robust cross platform system discovery script in POSIX. Using Autotools would not effectively change anything for end users at all and platform support would stay the same. Using `meson` would be a huge step backwards as far as reach goes.

> I think merging this one into mainline so people who want it can use it and then building on top makes sense. Thoughts?

Yes that makes sense. I'll go over this in a bit more depth and resolve the conflicts, then go ahead and merge for now and leave any configure tooling on the back burner for a while.

alerque:
As much as I *want* to get this merged and very much appreciate the work that went into it already, in this state it basically is trading one set of sins for another.

1. The BSD end of things is just a placeholder, it isn't actually implemented. I know this (because I can read comments! thanks for those) but I figured I could port the GNU stuff over myself. And maybe I could have, but, see №2
2. The `BSDmakefile` name hack doesn't actually work to catch BSD make variants. It failed on the first one I tried (NetBSD), and I read the documentation on OpenBSD only to see that it's going to have the same problem. I think there is a way out of this by using `Makefile` for the BSD one, a non-standard name for the include file, and trust `GNUmakefile` to catch GNU variants. I think it *could* be done, but I have also been down this road before and although I can't remember the exact pitfalls I ran into (old GNU make's being different perhaps? Or GNU make an macOS trying to imitate BSD make? Something wacky, I forgot) I do remember that I tried on a project for over a year and finally gave in and used Automake. And while it took a while to work out the GNU specific bits from my code, having it just work normally in the end was a relief.
3. Sans 1 and 2 being fixed, `bmake` currently does this aweful thing where the *first* thing it tries to do is delete `/usr/bin/vcsh`. Mind you I'm not even running the install target, it just finds that first and goes with it. Not pretty, and this makes any non-bullet-proof solutions to №1 and №2 hard to justify.
4. The automagical switching of what targets even get run when giving the same input bugs me. If I type `make all` and my buddy types `make all` I expect the same list of artifacts to be generated. Sometimes getting a man page and sometimes not is ... snazzy coding and bad user experience. At least in my personal opinion (as a distro packager) it's better to just have more specific commands that do what they say on the tin and error if they can't.
5. None of this clears the original list of sins, in fact many of them remain:

    * Test coupling I can fix easily (and did in my local branch of this).
	* Installing should be deterministic. Maybe a man page maybe not is not solving the Ruby problem.
	* ZSH completions are still installed where they don't belong. Just because one some other package put something the wrong place (including an older version of VCSH) doesn't mean that's where it goes. Also it doesn't account for the upstream ZSH packaging including something but wishing vendor somethings to be somewhere else.
	* Okay this one is fixed.
	* This is improved but not fixed. There are more hook points, but no discovery of the right standard directories even on systems that provide this info.
	* Fixed

I hope this doesn't come off the wrong way, I really do appreciate the effort that went into this, but I can't just patch it up with a few fixes and make it workable. It needs an overhaul. I'll work on that on my own branch, but I'll probably take pass at other solutions. In the mean time I may post some of the lighter weight fixes here that are portable separately.

Issue #168, opened 2015-06-15
"vcsh which" parameter check broken


tribut:
As discussed on IRC, since https://github.com/RichiH/vcsh/commit/d946b07817ffe6e156c159e7fc6270322413e530 `vcsh which` no longer accepts a substring as documented but expects the full path to a file that exists in the work tree.
The correct way to determine the exit code would probably be to remember if `grep` ever returned 0 (matched anything).

RichiH:
`sed` is [part of POSIX](http://pubs.opengroup.org/onlinepubs/009695399/utilities/sed.html) so the new external dependency is fine.

I really dislike that we would need another tempfile. I like the substring part, though.

All in all, wouldn't it be more versatile to see if there's any output from the loop and error out if that output is empty?

RichiH:
@tribut : I just cherry-picked your b3d7fb34868229489a1216f810a004089665bcbf and I think I will merge 9cf8f5c46b230abd912657aeac276e4a2ce38ce0 into master unless you object.

tribut:
Seems fine.

RichiH:
Thanks :)

PR #169, opened 2015-06-16
Speed up `vcsh which` and allow substring matching again

tribut:
Noticed that `which` was really slow, so I replaced the inner for-loop by a simple sed-replacement.
This PR also solves #168, albeit somewhat ugly:
 - `grep .` is called just to get a non-zero exit status when output is generated. This feels unnecessary, but `PIPESTATUS` is a bash/zsh-only feature and this is still orders of magnitudes faster than the loop
 - to get the exit status outside of the outer loop, I actually had to use a tempfile, because pipes spawn a subshell and cannot modify variables


RichiH:
Closing this. See #168 for details.

PR #179, opened 2015-08-20
Fix broken list-tracked-by -- list_tracked expects repository at $2

dottedmag:


RichiH:
Can you provide context for this patch, please? I don't see what it fixes.

dottedmag:
vcsh list-tracked-by foo
shows all tracked files, not the ones tracked by foo.

But I see that this command is deprecated, so you can ignore this patch.

RichiH:
Deprecated does not mean it's not worth fixing it. Thanks :)

PR #183, opened 2015-09-21
vcsh: Always fast forward initial merge

ohspite:
Ignore `merge.ff = no` if set in .gitconfig.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -169,7 +169,7 @@ clone() {
 	[ x"$VCSH_CONFLICT" = x'1' ]) &&
 		fatal "will stop after fetching and not try to merge!
   Once this situation has been resolved, run 'vcsh $VCSH_REPO_NAME pull' to finish cloning." 17
-	git merge origin/"$VCSH_BRANCH"
+	git -c merge.ff=true merge origin/"$VCSH_BRANCH"
 	hook post-merge
 	hook post-clone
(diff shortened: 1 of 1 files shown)


ohspite:
If `merge.ff = no ` is set in .gitconfig, then running `vcsh clone` results in

    fatal: Non-fast-forward commit does not make sense into an empty head

RichiH:
Can you expand with what went wrong? I don't have context for this PR.

ohspite:
Sure, thanks for looking at it.  

The `clone()` function in in vcsh does a `git init`, then configures the remote, then `fetch`, then `merge`. There's also some special vcsh stuff going on in the function, but all it's roughly analogous to a `git clone`.  

However, while `git clone` runs as one step and only looks at clone options in the user's global .gitconfig file, vcsh calls `git merge` separately and so git will use whatever merge options are set in the config file.  I have `merge.ff = no` in that file because that's how I like my trees to look.  And since you are merging into an empty head during a `vcsh clone`, git complains about a non-fast-forward commit and exits.  

Performing a fast-forward merge at that `clone` step is necessary and it can't mess anything up since it's the first commit, so forcing the command line option to do a fast-forward shouldn't create a problem for any users.

RichiH:
Thanks. Merged :)

PR #187, opened 2015-11-16
remove test from install dep, clarify README/INSTALL

nbirnel:
`make install` fails in master without a few perl modules; I have fixed this and put installation in the README TOC

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ all=test manpages
 
 all: $(all)
 
-install: all
+install: manpages
 	install -d $(DESTDIR)$(PREFIX)/bin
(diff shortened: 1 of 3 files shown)


RichiH:
Thanks for the patch. I suspect you might want to look at the release branch.

Can you split out into two patches? One for "get rid of the tests by default" for release and one for "all the doc updates" for release? Ideally, explaining the difference.

Yes, it would have taken me less time to do this myself than to write back, but I believe in attaching the correct name to commits ;)

nbirnel:
@RichiH Thanks for the feedback. I will separate the pull requests as you requested, but it will be a few days, as I am under the weather.

RichiH:
No worries. Thanks :)

PR #191, opened 2016-01-07
Fixed issues with cloning when using git version 1.7.1.

arndtc:
I was having trouble when cloning a new repo which using git version 1.7.1.
First trying to do a checkout kept giving an error from Git when using a checkout before the fetch.
  fatal: You are on a branch yet to be born

This issue doesn't happen on machines where I have access to newer versions of git, so I don't know when it was resolved for sure.

Second, the -c option wasn't added to git until version 1.7.2, so I just had to drop that option if using 1.7.1 or earlier.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -151,10 +151,24 @@ info() {
 }
 
 clone() {
+	GIT_VERSION=$(git --version)
+    GIT_VERSION_MAJOR=$(echo $GIT_VERSION | awk -F '[^0-9]*' '$0=$2')
+    GIT_VERSION_MINOR=$(echo $GIT_VERSION | awk 'BEGIN {FS="."} {print $2}')
+    GIT_VERSION_REV=$(echo $GIT_VERSION | awk 'BEGIN {FS="."} {print $3}')
 	hook pre-clone
 	init
 	git remote add origin "$GIT_REMOTE"
-	git checkout -b "$VCSH_BRANCH" || return $?
+    if [ 1 -lt "$GIT_VERSION_MAJOR" ];then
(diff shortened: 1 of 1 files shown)


RichiH:
Thanks.

I don't see what lines 161-171 in your patch are for, they all seem to do the same.

I suspect falling back to the clone implementation in 7161ee4adc3f462f20ab9b55404ecf6968f2753f and telling the user that they simply won't be able to clone into branches if they use a Git version that ancient is a safer default, though.

arndtc:
Thanks for reviewing this and for your feedback.  Sorry my shell coding isn't that great.  Lines 161-171 actually were trying to preserve the branch option for later versions and then just drop it if using git version 1.7.1 and older.  But my code didn't really handle all the situations and correctly inform the user that the branch pull was being ignored.  Now that you've pointed out a tag where this might have worked.

I tried implementation https://github.com/RichiH/vcsh/commit/7161ee4adc3f462f20ab9b55404ecf6968f2753f but it gave me shell error.  So I went back one prior release https://github.com/RichiH/vcsh/commit/37bbb350fbe77109d6f8c00f83e64335a7401769 and that did work.  So I'll just use that release on machines where I only have access to older versions of git.

I'll close out this request.  Also since my fork doesn't really solve any problems I've deleted it.

RichiH:
Off-hand, I don't see how using 37bbb35 helps here, but if it works for you and you only need to use a it version that old on one machine, that seems to solve your problem.

If this turns out to be a recurring issue, I will introduc special code to handle old Git versions.

PR #196, opened 2016-02-14
Fix write_gitignore for git version >= 2.7

### LINKED ISSUE #195: Git >= 2.7 changed the meaning of `!`, breaking write-ignore
Right now, write-ignore produces a file like this:
```
!/foo
!/foo/bar
!/foo/bar/baz.txt
```
But, as of 2.7, all files and directories in `/foo` and `/foo/bar` will now show up as untracked. It is completely sufficient to specify only the actual file:
```
!/foo/bar/baz.txt
```

https://git.kernel.org/cgit/git/git.git/tree/Documentation/RelNotes/2.7.0.txt

> 
 * Allow a later "!/abc/def" to override an earlier "/abc" that
   appears in the same .gitignore file to make it easier to express
   "everything in /abc directory is ignored, except for ...".

jotrk:
As of git version 2.7 it is no longer necessary to include parent
directories in the list of files not to be ignored.

This fixes #195.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -476,14 +476,21 @@ write_gitignore() {
 
 	use
 	cd "$VCSH_BASE" || fatal "could not enter '$VCSH_BASE'" 11
+	local GIT_VERSION=$(git --version)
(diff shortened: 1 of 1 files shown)


kwbr:
Works good for me.

RichiH:
Thanks :)

PR #198, opened 2016-03-26
Fix git version parsing when there are spaces

ohspite:
On my machine, before this patch:
```
> git --version
git version 2.8.0rc3
> vcsh write-gitignore config-mr
/home/me/usr/bin/vcsh: 479: local: 2.8.0.rc3: bad variable name
```


RichiH:
Thanks

Issue #180, opened 2015-08-23
Make .gitignore.d & gitattributes.d XDG compatible


nfarrar:
Vcsh currently reads the repo ignore & attribute files from `$HOME/.gitignore.d/$VCSH_REPO_NAME` and `$HOME/.gitattributes.d/$VCSH_REPO_NAME` (respectively), however I (personally) prefer to keep `$HOME` as uncluttered as possible and group application configurations into application-specific configuration directories (using the [XDG](http://standards.freedesktop.org/basedir-spec/basedir-spec-0.6.html)) specification defaults:
```
$XDG_CONFIG_HOME/vcsh/.gitignore.d/$VCSH_REPO_NAME
$XDG_CONFIG_HOME/vcsh/.gitattributes.d/$VCSH_REPO_NAME
```

Would it be possible to add support for these paths?

nfarrar:
Just checking back in. Any chance to get support for these? :)

kwbr:
I tested the patch provided by @miramir (#203). This works for me.

It would be nice of course, to have $XDG_CONFIG_HOME as the default for the variables for newly created vcsh-repositories.

hvhaugwitz:
What is the state of this request?

alerque:
See also my comments in #203, I would like to see this fixed too, although issues of migration are of course sensitive.

PR #203, opened 2016-06-12
Make path to .gitignore.d and .gitattributes.d configurable

miramir:
Fix issue #180

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -78,6 +78,8 @@ fi
 : ${VCSH_HOOK_D:="$XDG_CONFIG_HOME/vcsh/hooks-enabled"}
 : ${VCSH_OVERLAY_D:="$XDG_CONFIG_HOME/vcsh/overlays-enabled"}
 : ${VCSH_BASE:="$HOME"}
+: ${VCSH_GITIGNORE_D:="$VCSH_BASE/.gitignore.d"}
+: ${VCSH_GITATTRIBUTES_D:="$VCSH_BASE/.gitattributes.d"}
 : ${VCSH_GITIGNORE:=exact}
 : ${VCSH_GITATTRIBUTES:=none}
 : ${VCSH_WORKTREE:=absolute}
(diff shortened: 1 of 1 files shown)


kwbr:
Any chance to release a new vcsh with this patch applied?

alerque:
Hey @RichiH any chance I could get the short version on the reasoning behind this PR closure?

alerque:
I might have found it, [your comment here](https://github.com/RichiH/vcsh/issues/16#issuecomment-3252116) explains what you were probably thinking. Since it's kind of off topic there maybe humor me for a minute here, but I'm not convinced the logic is correct:

> Users may have $XDG_CONFIG_HOME set to different locations on different machines or several users may use one repo. vcsh could not handle this without softlinks or other hacks. Thus, $VCSH_BASE/.gitignore.d is the only place that makes sense. I could make this a config option, but I doubt this is useful and only serves for people to shoot their own feet.

Of course `$XDG_CONFIG_HOME` may vary from machine to machine, but if it does wouldn't using that variable in our default config by definition solve that? And if the user chooses to override that as a default location, wouldn't the fact that the config is stored in the git repo meta data solve this anyway? If `vcsh` finds the repo at all it would by definition find the related ignores and attributes no matter where they were configured to go, no?

RichiH:
This seems to a side effect of the rename from master to main; and it took me some time figure out that the hint was in the mouse-over of the disabled re-open button... Weird

image

RichiH:
OK, https://github.com/miramir/vcsh is deleted and it seems that verification was triggered by the rename. But, given how GitHub stores forks as one repo in the background, https://github.com/RichiH/vcsh/pull/203/commits/e4addc805b9f05eeaa4187da16ef5742be38d911 still exists

RichiH:
I will report this as a bug to GitHub support.

RichiH:
@alerque: I don't get email for my own "closing" of PRs and issues. Did you get any other close emails at around the same time? If yes, we need to review them all.

alerque:
Yes. There were 4 of them as I recall. Since *most* of them automatically reparented and just those 4 closed I thought at first they were ones you had just decided to ditch—although the lack of comments made me wonder. I didn't notice anything unique about the 4 but now that you point it out I didn't review the status of the downstream forks. I'll track down the other 3.

PR #205, opened 2016-06-19
Fix documentation for running git commands

### LINKED ISSUE #204: Confusing documentation on running git commands
The readme file mentions this command:
`vcsh vim remote add origin <remote>`

I searched for a command named remote in the manual but could not find one. The man page of `vcsh` says:

```
   repo gitcommand
          Shortcut to run vcsh on a repo. Will prepend git to command.
```

This documentation has multiple discrepancies:
- the command description says `gitcommand` whereas the description mentions `command`. I guess they both are referring to the same thing.
- It says shortcut to run `vcsh` on a repo whereas the command seems to be a git command rather than a vcsh command. So it is a shortcut run git commands rather than vcsh.

The documentation printed by `vcsh` command itself seems to be saner:
   `<repo> <git command> Shortcut to run git commands directly`

We should perhaps copy the same to man page as well.

harendra-kumar:
Fixes #204

### DIFF
--- a/doc/vcsh.1.ronn
+++ b/doc/vcsh.1.ronn
@@ -41,7 +41,7 @@ vcsh(1) - Version Control System for $HOME - multiple Git repositories in $HOME
 
 `vcsh` write-gitignore <repo>
 
-`vcsh` <repo> <git command>
+`vcsh` <repo> <gitcommand>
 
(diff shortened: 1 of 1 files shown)


RichiH:
Belated thanks.

PR #206, opened 2016-06-20
Fix hooks running git commands

harendra-kumar:
If a hook runs a git command the command will work on the VCSH git repository because of the GIT_DIR environment variable being set and exported. 

This change unsets the environment variables before calling the hooks and restores it again after the hook is done.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -249,11 +249,19 @@ git_dir_exists() {
 }
 
 hook() {
+	# Save and unset the GIT_DIR environment variable so that it does not
+	# interfere with the hook scripts running git commands.
+	local VCSH_GIT_DIR=$GIT_DIR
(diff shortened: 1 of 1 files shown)


harendra-kumar:
It seems this is not going to work as the GIT_DIR env variable is being used by hooks in interesting ways e.g. https://github.com/vdemeester/vcsh-home/blob/bootstrap/bootstrap.sh uses it to set sparse checkout. Though it is not clear from the documentation whether this is officially exposed for hooks.

Can we instead export something like VCSH_GIT_DIR instead of GIT_DIR for the hooks to use officially? Though it will still break any existing hooks using it.

lyda:
I think #214 addresses this issue as well?

PR #210, opened 2016-07-11
Use git read-tree (dry run) to detect conflicts

harendra-kumar:
ls-tree does not account for sparse checkout settings and therefore
reports spurious conflicts if an existing file in the worktree conflicts
with a file in the repository which is not even being checked out. On the
other hand, read-tree takes care of sparse checkout settings.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -169,13 +169,8 @@ clone() {
 		git fetch origin
 	fi
 	hook pre-merge
-	git ls-tree -r --name-only origin/"$VCSH_BRANCH" | (while read object; do
-		[ -e "$object" ] &&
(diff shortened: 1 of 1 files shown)


alerque (changes requested):
Thanks for taking the time to contribute! I *think* this is the right thing to do, but I think it's probably something we have a test case setup to catch. Do you happen to still have an interest in this?

alerque:
This PR was merged as 22099c4. Because the fork repository is using GitHub's legacy system it wouldn't let me push the merge conflicts to the PR branch first, but I think everything got where it belongs. If not please holler!

PR #211, opened 2016-07-11
Fail the command if a hook script fails

harendra-kumar:
Currently vcsh commands succeed even if a hook script fails. This is not
the right behavior in my opinion. The user should know when a hook fails
rather than silently continuing even though it may not make sense.


alerque (changes requested):
This is great, but we don't need a new `die()` function, there is already a `fatal()` that does the same job. This should be refactored with that, but otherwise yes...

PR #216, opened 2016-11-17
Add a regular git commit alias.

lyda:
Git users coming from svn, hg, cvs and many other places
use "ci" as an alias to "commit."  I'd note that such users
are also used to "co" being an alias for "checkout" so the
"co" alias is kind of dangerous.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -529,7 +529,7 @@ VCSH_COMMAND=$1; export VCSH_COMMAND
 
 case $VCSH_COMMAND in
 	clon|clo|cl) VCSH_COMMAND=clone;;
-	commi|comm|com|co) VCSH_COMMAND=commit;;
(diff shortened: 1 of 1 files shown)


alerque:
I thoroughly agree that `co` is a bad alias for `commit`. Not only is it contrary to what that command is in many other source control systems, but a log of people have git aliases for it to be `checkout` in git as well. It might be wise to drop that one before somebody gets burned.

lyda:
I've done another version of this in PR #238 that retains the "co" alias. So pick whichever (or none) suits. I do think this one would be better though.

alerque:
As much as I agree with my previous self that `co` is a bad alias, at this point removing it is probably more disruptive than leaving it. On that topic I think something like #241  proposes with user control over *all* aliases is probably a better future direction that pruning existing aliases.

As for this PR, it has been obsoleted by #238 (just merged).

PR #219, opened 2017-04-26
Add section on submodules

gliptak:
https://github.com/RichiH/vcsh/issues/94

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,11 @@ Let's say you want to version control your `vim` configuration:
 If all that looks a _lot_ like standard `git`, that's no coincidence; it's
 a design feature.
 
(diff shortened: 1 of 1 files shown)


RichiH:
@gliptak did you try this? I broke horribly in the past

gliptak:
I tried it when proposed, and it seemed to work. Now retrying, I do see an error:

```
fatal: ambiguous argument 'origin/master..': unknown revision or path not in the working tree.
```

RichiH:
That seems to be the one from back then :(

I did talk to the maintainers of submodule about this in 2013 (I think) as it's something why would need to fix on their side...

PR #220, opened 2017-05-19
New testsuite based on Git's test-lib.sh

djpohly:
I've taken the tests that were written using Bats and ported them to Git's testing harness.  Advantages:
* Testing harness used for Git itself
* Harness written entirely in shell (removes dependency on Bats or Perl)
* Harness provides repository creation commands (removes need for vcsh_testrepo, which IIRC was the sticking point for merging back in February)
* Runs faster (on my box at least)

I'll keep tinkering and expanding but wanted to let you know this was ready and available.


djpohly:
:bulb: if you do merge, remember to fix the image URL for the Travis status in README.md

djpohly:
@RichiH Still have interest in this?  The current 165 tests are mergeable/usable/passing as-is (with 17 marked as expect_failure for existing issues).

alerque:
I'm a bit late to the party here so excuse my confusion, but can I ask what the motivation for this is? I could see wishing to move away from Perl, but why would some mataure test framework like `bats` be a detriment as a developer/CI dependency? If anything I would have looked into migrating **to** `bats` not away from it.

@djpohly Are you still interested in this? I don't want work to go to waste but I also want something that's actually going to be a long term benefit and at first blush I'm unclear on what the gain is.

djpohly:
There's no migrating to *or* away from Bats here; either testsuite is better than what vcsh currently has, which is essentially none.  You'll notice I wrote both - I was simply trying a couple of options for more thorough testing to see if one worked better than the other, and I happened to try Bats first.

In terms of long-term benefit, however, I found it much easier to write solid tests with Git's test-lib.sh harness.  Bats is pretty limited in what it does, and test-lib.sh is more feature-rich and mature, having been actively developed since Git 1.0 (2005).  Convenience functions for working with Git repositories, like `test_create_repo` and `test_commit`, were especially helpful when trying to write proper, self-contained tests rather than depending on an [existing remote repository](RichiH/vcsh_testrepo) to which you only have read-only access. Features like `test_expect_failure` are also generally useful for good testing practice, so that you can write a test immediately without having to wait for the feature to be implemented or bug to be fixed.

I'm sure that some of the relevant conveniences could be copied from test-lib.sh to a Bats library if that's what the devs want.  Or they can throw it all out.  Personally, it doesn't matter to me - I wrote tests so that I could potentially start working on refactoring vcsh, but since nothing was merged I moved on to other things.

alerque:
Thanks @djpohly that's helpful background for reviewing this.

djpohly:
Looks like there's a standalone harness based on Git's test-lib now: https://felipec.github.io/sharness/.  This testsuite could potentially be ported over if that would be preferable.

djpohly:
It's probably smart to squash-merge this if you don't need the whole history of trying different harnesses. :)

PR #223, opened 2017-07-08
Fix parsing of Git version numbers

madduck:
The Git minor version now has two digits, so fix the regexp parsing it.
We'll also anticipate the future and handle the time when the major
version hits 10 and parse accordingly

Signed-off-by: martin f. krafft <madduck@madduck.net>


RichiH:
Thanks!

PR #224, opened 2017-07-16
completion: Return the correct error code from __vcsh_*.

danielshahaf:


RichiH:
I will simply trust you on this one :)

PR #226, opened 2017-07-16
Add 'ls' short command

danielshahaf:


RichiH:
I am wondering if there will ever be an actual `ls` subcommand; can't say for sure, but introducing that at any time would then be a breaking change.

danielshahaf:
> I am wondering if there will ever be an actual `ls` subcommand; can't say for sure, but introducing that at any time would then be a breaking change.

Fair enough.  Thanks for merging the others!

alerque:
I'm not sure this is a good idea either. I love short aliases, but I *really* like to roll my own and am constantly bothered by upstream projects trying to add them for me.

This one in particular seems like it could be ambiguous in the future. What are we listing? Repositories? Tracked files? Untracked files? Some other scope? I'd rather not have to deal with removing this later and I don't want to document why this is so different that plain `ls`, `git ls-files`, etc.

Thanks for taking the time to contribute though, having extra eyes on what could be improved is appreciated.

PR #227, opened 2017-07-16
completion: Set the context correctly in 'foreach'.

danielshahaf:
This patch changes the zstyle context from `:completion::complete:git:…` to `:completion::complete:vcsh-foreach:…` when completing at `vcsh foreach <TAB>`.

The rationale is that `_vcsh` shouldn't be looking for completions under the `*:git:*` namespace.  The command being run is vcsh, not git.

The second hunk will have a trivial conflict with #225.  They change adjacent lines, but they're semantically independent of each other.

### DIFF
--- a/_vcsh
+++ b/_vcsh
@@ -24,7 +24,7 @@ function _vcsh-enter () {
 }
 
 function _vcsh-foreach () {
-	_dispatch git git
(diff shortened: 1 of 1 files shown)


RichiH:
Thanks!

PR #228, opened 2017-07-17
Do not use shared Git repositories

ttuegel:
Shared repositories were created using `git init --shared=0600` with the intent
of keeping configuration data private. Git reports an error if "shared"
repositories are created with private permissions. (Due to an apparent bug, the
error was not reported before git-2.13.2.)

Instead of creating a shared repository, use `umask 0077` to make created files
accessible only to the current user. The umask setting is inherited by child
processes and respected by Git.


RichiH:
Thanks!

PR #232, opened 2017-08-23
Add bash-completion

### LINKED ISSUE #170: add bash_completion functionality
It would be great if vcsh would include a bash_completion script, which the user could plug into `/etc/bash_completion.d`.  For example, I could then type

```
vcsh e <TAB>
```

and have 

```
vcsh emacs
```

in my terminal.  This would save quite some typing.

rhopfer:
Add full bash-completion including git sub commands (close #170).

### DIFF
--- a/_vcsh_bash
+++ b/_vcsh_bash
@@ -1,0 +1,138 @@
+# bash completion for vcsh.
+
+# run git command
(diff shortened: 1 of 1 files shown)


RichiH:
Thanks!

PR #233, opened 2017-08-27
Fix: list-tracked in subdir would print invalid paths

### LINKED ISSUE #230: list-tracked outputs invalid paths

rhopfer:
Fixes #230

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -274,7 +274,7 @@ list() {
 
 get_files() {
 	GIT_DIR=$VCSH_REPO_D/$VCSH_REPO_NAME.git; export GIT_DIR
-	git ls-files
+	git ls-files --full-name
 }
 
 list_tracked() {


RichiH:
I suspect I should have a look at all other ls-files, as well. I dimly remember not being able to fix an issue with gitignores...

RichiH:
(That does not mean I would stop merging your PRs, mind)

danielshahaf:
> I suspect I should have a look at all other ls-files, as well. I dimly remember not being able to fix an issue with gitignores...

The other callsites of `ls-files` all do either `cd "$VCSH_BASE"` or `export GIT_DIR`, so should be okay.  (Okay for now, I mean; a future patch might accidentally break this.)

RichiH:
Do you see any hurt in just doing it in all the places? I can't think of any.

danielshahaf:
> Do you see any hurt in just doing it in all the places? I can't think of any.

On the contrary: I think that changes to improve the code's robustness --- e.g., to ensure that GIT_DIR is always set --- are the way forward.

I also wonder what could be done to ensure that _future_ subcommands don't make this mistake.  For example, perhaps there should be a run_git() { env GIT_DIR=... git "$@" } helper function?

danielshahaf:
Speaking of robustness: delete_files() uses 'echo "$file"' which isn't well-defined when $file starts with a minus or contains a backslash.

PR #234, opened 2017-08-28
completion: Complete only external command for 'run'.

danielshahaf:
The difference is easiest to see when the following style is in effect:
```
zstyle ':completion:*' group-name ''
```

This prevents aliases, functions, names of shell paramaeters, etc from being offered.


RichiH:
Thanks!

PR #235, opened 2017-08-28
Number of bug fixes I use at home.

lyda:
Note I already have a PR or two with these fixes. This is just a snapshot of what I currently use. Feel free to merge all or part of it. Or none of it. Whatever.

* Fix parsing of git versions on OS X.
* Fix hook handling.
* Add editor modeline blurb to follow file standards.
* Do not abbriviate commit as co; do abbreviate as ci.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -166,7 +166,7 @@ clone() {
   You should add files to your new repository."
 		exit
 	fi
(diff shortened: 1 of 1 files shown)


RichiH:
Can you please split them into singular PRs?

PR #237, opened 2017-08-29
Add editor modeline.

lyda:
This will make sure people have a better chance of following the
code style of this file.

Note: I use `expandtabs` in vim since most of the world hates tabs. I don't though and it appears neither do you so this makes sure my default settings don't mess up this file.


rhopfer:
I don't hate tabs and most of the code in this file is indented by tabs.

lyda:
Yes, I get that. The modeline exists to configure editors to keep tabs in use. It's there to help other contributors maintain the current indentation style.

rhopfer:
Ah, I misread it, sorry. Than I am not happy with the vi modeline, because I prefer four spaces for tabs.
I suggest just to use 
```
# vi: noet:
```
for vi modeline.

lyda:
Good call. Done.

alerque:
@lyda and @rhopfer how would you guys feel if I were to replace this with an EditorConfig config file at the project level instead? Clearly the mixed styleing in the code means this could use addressing but I think EditoryConfig is a bit more ubiquitous than this covers ... on the other hand it doesn't catch a couple of the basics this does without plugins. Thoughts?

lyda:
I use vim and have editorconfig support in it - makes things way easier!

alerque:
Okay then. I suggest we setup a project level EditorConfig and leave the modelines out of it. We can also add a CI linter to make sure PR's don't introduce cruft so people that aren't setup to conform to editor config "hints" won't need extra tooling.

alerque:
I just tried to set this up in EditorConfig, but ran into a bug I've never noticed before in the Vim plugin ([reported here](https://github.com/editorconfig/editorconfig/issues/444)). I still think this is the way to go, but meh.

alerque (approved):
@lyda Does this still look good to you? I preserved the modeline variant in history, but switched to a project level Editor Config file. If people don't have it that's okay, they don't need it unless they want it. I also added a CI job to lint the style so we can see at a glance if PRs are introducing a mess. Does this work for you still? I kind of hijacked your PR ;-)

lyda:
LGTM.

While we're linting, might want to look into [shellcheck].

[shellcheck]: https://www.shellcheck.net/

alerque:
Yup yup. See #279 for my comments on that. I figured it can go in the lint workflow as another job, but I want to deal with some PR reviews first.

PR #238, opened 2017-08-29
Alias commit to ci as well.

lyda:
Mercurial, subversion and a common git alias for commit is "ci".
The PR adds that alias. Note that all of those also alias checkout
as "co" so this current case statement is a bit confusing.  That
said it doesn't make a huge amount of sense to do vcsh checkout so
it's fine if this stays if people have it in muscle memory.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -534,7 +534,7 @@ VCSH_COMMAND=$1; export VCSH_COMMAND
 
 case $VCSH_COMMAND in
 	clon|clo|cl) VCSH_COMMAND=clone;;
-	commi|comm|com|co) VCSH_COMMAND=commit;;
+	commi|comm|com|co|ci) VCSH_COMMAND=commit;;
 	delet|dele|del|de) VCSH_COMMAND=delete;;
 	ente|ent|en) VCSH_COMMAND=enter;;
 	hel|he) VCSH_COMMAND=help;;


rhopfer:
What do you think about a general functionality to add aliases in a config file, like Git does?

lyda:
I'm a fan. Though you start to push the boundary of what shell scripts can/should do.

rhopfer:
I created the issue #240 for that. I have time next days and will try to implement it.

rhopfer:
Done. @lyda Can you please have a look on [it](https://github.com/rhopfer/vcsh/tree/feature-aliases)?

lyda:
Note: this is not my project so it's @RichiH's opinion that matters.

But since I was asked...

Wow, very clean. I like it. I'd make one suggestion - remove the case statement with the old_alias stuff and that error message. Instead if vcsh starts up without an alias file it should make a canned one with all the old aliases. A little comment at top saying not to delete the alias file entirely.

lyda:
Also you could use `git` for mucking with the aliases. Assuming the alias file (or maybe the config file) looks like this:

```ini
[alias]
        ci = commit
        rm = delete
```

Then you can just do:

```bash
# Get an alias
git config -f $XDG_CONFIG_HOME/vcsh/aliases --get alias.ci
# Set an alias
git config -f $XDG_CONFIG_HOME/vcsh/aliases alias.ci commit
# Remove an alias
git config -f $XDG_CONFIG_HOME/vcsh/aliases --unset alias.ci
```

lyda:
Also you could just use `git config --global` and use the user's `~/.gitconfig` file and just use something like `vcsh-aliases` to store vcsh aliases.

Though detection of an initial run to pre-populate aliases might be harder then.

rhopfer:
> Note: this is not my project
I know, but you were interested and because you are not watching the project I thought I give you a ping.

I don't really like the idea to automatically create an aliases file. There are users who don't need them and you must opt-out aliases you don't like.

alerque (approved):
See also #226 for similar but subtly different rejected `ls` alias. In the case of `ci` there are so many VCS systems that use that alias already I can't picture ever wanting to use it for anything else.

That being said I'm interested in what we can do with #241 and moving alias control over to end users.

Thanks for the contribution and the discussion.

Issue #240, opened 2017-08-29
Add functionality to define aliases


rhopfer:
Instead of hard coded aliases allow their definition in ```/etc/vcsh/aliases``` like git does, e.g.
```
ci = commit
co = upgrade
```

PR #241, opened 2017-08-29
Feature aliases

rhopfer:
My proposal to add dynamic aliases (see #240). Please check the documentation I wrote because my english is bad. Suggestions are welcome.

### DIFF
--- a/doc/vcsh.1.ronn
+++ b/doc/vcsh.1.ronn
@@ -5,7 +5,7 @@ vcsh(1) - Version Control System for $HOME - multiple Git repositories in $HOME
 
 `vcsh` [<options>] <command>
 
-`vcsh` alias [<alias-name>]
+`vcsh` alias [<alias>[=<command>]]
 
 `vcsh` clone [-b <branch>] <url> [<repo>]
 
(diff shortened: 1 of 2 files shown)


alerque:
Thanks for taking the time to contribute this yet. Overall the concept looks really interesting and I would even suggest that (if we go this route) the existing predefined aliases should be dropped in favor of a default alias config file that would be superseded by a user config if any. I have some concerns that the regex based approach to managing the content of the aliases file might be a bit complex and possibly brittle.

That being said before I review this too deeply or we make a decision I have to ask ... did you consider storing aliases *in Git's config system*? It's possible to set and retrieve arbitrary keys in there and since a working Git config is already an absolute requirement that might be a way to avoid the overhead of having a separate alias file handling system. Not that we're afraid of more dotfiles of course ;-)

alerque:
Just for reference, when reviewing this lets compare what GitHub did with their official [CLI](https://github.com/cli/cli/) and its `gh alias`.

RichiH:
Also Git's; I suspect everyone will know this, but just to make sure: https://git-scm.com/book/en/v2/Git-Basics-Git-Aliases

alerque:
Just in case it turns out useful later, it seems this PR's (currently only) commit is a squash of all the bits in [this branch](https://github.com/lyda/vcsh/tree/feature-aliases).

PR #245, opened 2017-11-28
Make <repo> arg to write-gitignore optional

madduck:
If we're within a repository, it seems silly to expect the repository
name to be specified to write-gitignore. So this patch makes it that the
argument is optional if we're within an existing VCSH checkout. You can
still explicitly provide $2 to override the repository for which to
write the gitignore file.

Signed-off-by: martin f. krafft <madduck@madduck.net>


alerque (approved):
Thanks for this contribution @madduck! I've wanted this so many times...

There are a couple of code style problems (you mixed tabs and spaces) but I'm going to send it through anyway because the *rest* of the script has the same problem elsewhere and I'll cleanup some of the style issues in a single pass later.

RichiH:
Tabs for indenting, whitespace for formatting; not sure if that style is still followed much these days, though.

PR #250, opened 2018-04-15
Fix myrepos status in vcsh repositories

aryelgois:
In [myrepos](http://myrepos.branchable.com/), when running `mr status` with repositories maintained by vcsh, files are listed with a relative path from the git repository (usually inside `~/.config/vcsh/repo.d/`). As it is common to be at `$HOME` when doing so, files are prepended with `../../../../`

That really bothers me.

I looked at the source code of both repositories and found that it was not always like that. Before using vcsh `status` function, myrepos was disabling the git config `status.relativePaths`: RichiH/myrepos@c6a37fa0d2309fe6a16948ec33b7ad78899dec53.

So I recreated the old behavior in `status_helper`.

Maybe it was too extreme, maybe it was actually useful. And it is not the only solution: before setting that config, myrepos was changing the directory to `core.worktree`: RichiH/myrepos@eabad02ab25169b9658d87fbdebd32a3ac86b317

This issue could be fixed directly in myrepos, because it does not affect `vcsh status`, by reverting to the older implementation:

<pre><code>vcsh_status = <strong>cd $(vcsh run "$MR_REPO" git config --get core.worktree);</strong> vcsh status "$MR_REPO" "$@" || true</code></pre>

I didn't test, but relativePaths config seems more efficient.

> also the github repository for myrepos is outdated, and I couln't find a way to make a 'pull request' in http://myrepos.branchable.com


alerque (commented):
Thanks for taking the time to contribute this. I did some testing on my system and it works great and looks much better.

My only concern is that it no longer shows any indication that the path is relative to `$HOME` instead of `$PWD`. As a long time user this is intuitive to me but I'm worried it might be confusing — perhaps there is a way to prefix the paths with `~` or similar? Do you know if there is anything we can do about that? Or maybe we should persue fixing this on the `mr` end of things. Thoughts?

alerque:
This won't be quite as easy to fix up as I was hoping, at least not with out loosing a feature: Git's automatic color output that can be user configured. In order to add this prefix we'd have to munge the output from `git status`. As soon as you pipe it to something color turns itself off. We can work around that, but then you have the difficulty of matching and doing anything sensible and robust. Using `sed` for this is a bit awkward, `awk` would be better suited but that's a new dependency. This sort of works in shell, but I have no idea how brittle it might be:

```patch
--- a/vcsh
+++ b/vcsh
@@ -430,7 +430,10 @@ status_helper() {
 		[ ${commits_behind} -ne 0 ] && echo "Behind $remote_tracking_branch by $commits_behind commits"
 		[ ${commits_ahead} -ne 0 ] && echo "Ahead of $remote_tracking_branch by $commits_ahead commits"
 	}
-	git ${VCSH_GIT_OPTIONS} status --short --untracked-files='no'
+	( git ${VCSH_GIT_OPTIONS} status --short --untracked-files='no' ) |
+		while read status path; do
+			echo " $status ~/$path"
+		done
 	VCSH_COMMAND_RETURN_CODE=$?
 }
 ```

 I need sleep. Maybe a new day's coffe will suggest a cleaner fix.

alerque:
I don't think there is a way to do this cleanly. The output of [`git-status`](https://git-scm.com/docs/git-status) is not very configurable. In particular there is no `--format=` override like there is for `git-log` that would allow us to match the original format with escape codes and all plus inject our string.

It does have a porcelain output option that would be the right way to fetch the data if we were going to parse and amend it ourselves. This would be the *best* way to modify the path to anchor it to `$HOME`. The problem is we don't really want machine readable output, we want user readable output. We could reinvent the wheel and add our own color coding and pretty formatting back in, but that's a mess. I for one really want the output to respect my color settings and rely on the at-a-glance info red/yellow/green provides.

Alternatively we can muck around with the human readable output, possible color escape sequences at all. Whether there are control characters with color data or not in the stream it seems really consistent that the location we need to insert our anchor is after space separator. Unfortunately space is also used to left pad the first field. In the event of a merge status something simple like `sed -E 's@ @ ~/@2'` would fail.

The best compromise I could find is what's now in this PR. Output on my system looks like:

image

alerque (approved):

alerque:
I'm kind of baffled as to why the tests are failing in CI, they work for me locally, the failing test has nothing to do with the change, and the same test works on *main*.

RichiH:
It might be related to https://stackoverflow.com/questions/44101948/prove-returns-inconsistent-test-results-with-testmore-and-tap-extension ; I bumped my time-box for vcsh (trying to get back in at a long-term pace), but wanted to document it here.

alerque:
Hm, interesting lead, but that was supposedly fixed in Test::Harness: 3.41. The GitHub Action downloaded and installed 3.42. Also we're not using `.tap` but `.t`.

RichiH:
While it's known-flaky, I am fine with merging while tests pass for you
locally. Not ideal, but...

>

Issue #247, opened 2018-02-17
foreach greppable mode


danielshahaf:
I'd like the `foreach` command to be able to prefix the repository name to each line of the output.  This will allow, for example, `vcsh foreach ls-files ./ | egrep '^(foo|bar)'` to show all files in the current directory that belong to either the 'foo' vcsh repository or the 'bar' vcsh repository.

WDYT?

RichiH:
I like the idea, but this would probably mean catching all output and printing it ourselves. Does anyone have any other ideas?

danielshahaf:
Bueller? Bueller?

julien-lecomte:
#252 

My solution is just piping the output to `cat` if there's nothing to do, and to `sed` otherwise.

I've changed another behavior: before a single line would be output containing the repo name. This seems unnecessary now with the new option, and even a bit unwanted.

--help was updated, no additional unit tests were added.

danielshahaf:
Thanks @julien-lecomte :)

alerque:
The desired usage looks like it's doing more work than it should be:

```console
$ vcsh foreach ls-files ./ | egrep '^(foo|bar)'
```

What are you going to do with these results that would not be better served by just going straight for the data? Like this perhaps:

```console
$ xargs -n1 vcsh list-tracked <<< "foo bar"
```

This will avoid iterating over repositories you don't want results from anyway.

If what you want is the name of the repo instead of the names of files, why not use `vcsh which`?

I'm not saying there are no use cases for adding a prefix, but it would be nice to have a use case that isn't already better served by some other command. That would make any actual implementations easier to evaluate.

danielshahaf:
@alerque If I am interested in specific repos called `foo` and `bar`, then yes, that xargs idiom would be good (assuming the end of the output for `foo` / start of the output for `bar` can be recognized).

I think my original use-case was to eyeball the output, and I was hoping for a more SQL-like output mode: "first column: repo name, second column: whatever the `foreach`'d command printed".

Having the repo name on each line would also allow sorting the `vcsh foreach ls-files` output by filename rather than by repo name, while still being able to track, for each file, what repo it is in.

In general, output is easier to parse if all lines are of the same form.  Right now, some lines are `${repo_name}:` and some are output of the command for that repo, so parsing/post-processing has to account for these different kinds of lines, keep state, etc..

RichiH:
Putting it differently: The data is currently somewhat normalized in the form of
`name1`
`output1`
`name2`
[...]

PR #252, opened 2018-06-18
Enable vcsh ls-file to take an option '-p'

julien-lecomte:
Enable vcsh ls-file to take an option '-p' to prefix all lines with repo name

cf: #247

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -110,7 +110,8 @@ help() {
    delete <repo>        Delete an existing repository
    enter <repo>         Enter repository; spawn new instance of \$SHELL
                         with \$GIT_DIR set.
-   foreach [<-g>]
+   foreach \\
+     [<-g>] [<-p>] \\
      <git command>      Execute a command for every repository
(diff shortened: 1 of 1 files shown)


vcsh
```diff
@@ -234,17 +235,20 @@ foreach() {
...
-	while getopts "g" flag; do
+	command_suffix='cat'
+	while getopts "gp" flag; do
 		if [ x"$1" = x'-g' ]; then
 			unset command_prefix
 		fi
+		if [ x"$1" = x'-p' ]; then
+			command_suffix='sed "s/^/$VCSH_REPO_NAME: /"'
+		fi
 		shift 1
```
danielshahaf: Does that DTRT when argv is `-gp` as one word?

julien-lecomte: No, and it's a limitation of the `shift 1` approach.

The getopts $flag variable should be parsed and then handled with a case/esac.

PR #253, opened 2018-10-28
Don't try to push/pull from repos without remote

tikki:
I have some repos managed by vcsh that are local only (e.g. one for my ~/.ssh dir). When I run `vcsh push` or `vcsh pull` I get these annoying spam'y error messages from git that I need to set up tracking information, which is fine but unnecessary.

This PR changes that so vcsh only tries to pull/push repos that have a remote set.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -276,6 +276,13 @@ list() {
 	done
 }
 
+list_has_remote() {
+	for VCSH_REPO_NAME in $(list); do
+		GIT_DIR=$VCSH_REPO_D/$VCSH_REPO_NAME.git; export GIT_DIR
+		git config remote.origin.url >/dev/null && echo "$VCSH_REPO_NAME"
+	done
+}
+
 get_files() {
 	GIT_DIR=$VCSH_REPO_D/$VCSH_REPO_NAME.git; export GIT_DIR
 	git ls-files --full-name
@@ -354,7 +361,7 @@ list_untracked_helper() {
 
 pull() {
 	hook pre-pull
-	for VCSH_REPO_NAME in $(list); do
+	for VCSH_REPO_NAME in $(list_has_remote); do
 		printf '%s: ' "$VCSH_REPO_NAME"
 		GIT_DIR=$VCSH_REPO_D/$VCSH_REPO_NAME.git; export GIT_DIR
 		use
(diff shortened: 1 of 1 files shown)


vcsh
```diff
@@ -277,6 +277,13 @@ list() {
 	done
 }
 
+list_where_remote() {
+	for VCSH_REPO_NAME in $(list); do
+		GIT_DIR=$VCSH_REPO_D/$VCSH_REPO_NAME.git; export GIT_DIR
+		git config remote.origin.url >/dev/null && echo "$VCSH_REPO_NAME"
```
alerque, line 283: What if `origin` isn't the remote? I realize `vcsh clone` sets that up by default but I have some systems where it's been changed, this would cause basic functions to fail. Right now the only place where that is hard coded is `clone()` and I think it's probably useful to keep it that way.

alerque (changes requested):
Thanks for taking the time to contribute! This looks useful, and I've certainly been spammed with messages before because I sometimes keep local config repos around too. But there is one nit pick I'm not sure what to do with off the top of my head. Any ideas what to do about repos that do have a default remote but it isn't `origin`?

vcsh
tikki: good point, I changed it to simply check for any existing remote. whatever the repo has set up as default remote should be used then, and if there's no remote it'll be ignored by vcsh's push operation.

vcsh
```diff
@@ -276,6 +276,13 @@ list() {
 	done
 }
 
+list_where_remote() {
+	for VCSH_REPO_NAME in $(list); do
+		GIT_DIR=$VCSH_REPO_D/$VCSH_REPO_NAME.git; export GIT_DIR
+		[ -n "$(git remote)" ] && echo "$VCSH_REPO_NAME"
```
alerque, line 282:
```suggestion
		git config branch.$VCSH_BRANCH.remote > /dev/null && echo "$VCSH_REPO_NAME"
```

How about this to check if there is a remote for the branch VCSH thinks it is operating on? This way we can also use the return code of `git config` rather than checking whether STDOUT exists (which by the way might leak expected bits of STDERR to the user).

alerque (changes requested):
Minor nitpick, but I think we're getting there. Thanks again for the contribution and for polishing it up.

alerque (approved):

vcsh
```diff
@@ -276,6 +276,13 @@ list() {
 	done
 }
 
+list_where_remote() {
```
RichiH, line 279: Minor nit: `list_*` follows list, verb, optional. As such, could you rename to `list_has_remote`, please?

RichiH (changes requested):
Minor nit: `list_*` follows list, verb, optional. As such, could you rename to `list_has_remote`, please?

Else, it's good to merge, thanks!

alerque (approved):

PR #254, opened 2018-10-28
Add an option for `push` and `pull` to run their subcommands in parallel

tikki:
I usually have 3-6 repos on each machine that are managed by vcsh. When running `vcsh push` or `vcsh pull` it will go through all of them one after the other, even when most of them have no changes. This is bothersome and makes vcsh feel unnecessarily slow.

I propose adding a `-j` option to `vcsh push` and `vcsh pull` which runs all of the git commands as background subprocesses in parallel. The option name is based on `make`'s `-j` option which does roughly the same, but smarter and with the option to set the number of processes.

I made it an option and don't default to it, because the way it's implemented right now we lose `git`'s standard output. Having all sub-processes run at once and output to stdout, we'd end up with a garbled mess, so I opted to add `--quiet` to the call to `git`.
One way to handle this would be to redirect the subprocess' output to temporary files and display them each once they finished. This could easily be done in the `wait` loop.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -120,8 +120,8 @@ help() {
    list-untracked \\
         [<-a>] [<-r>]
         [<repo>]        List all files not tracked by all or one repositories
-   pull                 Pull from all vcsh remotes
-   push                 Push to vcsh remotes
+   pull [-j]            Pull from all vcsh remotes
+   push [-j]            Push to vcsh remotes
    rename <repo> \\
           <newname>     Rename repository
    run <repo> \\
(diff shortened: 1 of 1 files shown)


alerque:
Running these jobs in parallel is certainly a huge speed boon, but can I suggest that `mr` might be better positioned to provide this? It already handles the job runners and collates output without mixing or discarding the streams. Is there something to be gained from `vcsh` providing this directly? And if not `mr`, why not GNU `parallel`? If there is some reason this has to be baked in to VCSH I'm willing to consider it, but parallel job runners are **hard** and I think it would behoove us to shove that burden off on some other project if we can.

RichiH:
myrepos has had native vcsh support for ages now; it's detected and handled correctly. @tikki could you try this and see if it covers your use case?

tikki:
I drastically reduced the amount of repos I manage using vcsh, so this feature is no longer necessary for me.

I think I tried adding the feature to vcsh and not use a separate tool because I didn't want to require separate tools to be installed to achieve this goal. I chose vcsh in the first place because it's a (somewhat) simple and portable shell script.
I also understand if subprocess management seems a bit out of scope for vcsh though.

Anyway, like I said, I don't need the feature anymore so from my side we can simply close this PR.

alerque:
Thanks for the response @tikki.

For others I think this is where we are at:

* Doing parallel job management is hard. Not impossible, but hard to do right.
* There are already at *least* two tools that do it well in combination with `vcsh`: `mr` specific to bulk repository management and `parallel` for general purpose execution.
* If you have only one or a few VCSH repos then this isn't a huge felt need and probably not worth the development effort and risk of doing it wrong and introducing bugs.
* If you have a lot of VCSH repos, we **strongly** recommend `mr` anyway.
* If after all those considerations somebody can make a use case for why this is needed we can re-evaluate, but for now the answer is we have to plan to implement this.

PR #115, opened 2014-03-07
Agnostic

andschwa:
As per PR #114, this adds the `vcsh setup` and `vcsh where` commands, with necessary documentation. It also enables the sourcing of `./.config/vcsh/config`, which `vcsh setup` will automatically create with the `$VCSH_BASE` and `$XDG_CONFIG_HOME` variables set to `$(pwd)` and `$(pwd)/.config` respectively, thus enable "agnostic" use of `vcsh` from the folders with the a `./.config/vcsh/config` file.

It's a work in progress, and probably needs more testing, but I think it's a good way to go about doing this.

Issue #110, opened 2014-03-02
How to manage /etc files


a-sk:
Hello.

Could you guys, tell me how to manage `/etc` files with vcsh, I didn't find any clean solution.

alerque:
I'd be happy to review work that make `vcsh` more flexible and robust at dealing with non-HOME base points. To this end I've re-opened #115 as #283 for further review.

For my own part I use `etckeeper` for /etc and am quite happy with it, but I can see the use case for some people and could see this being useful for other reasons.

If anybody else is still actively interested in working on this please feel free to jump in.

RichiH:
CC @bremner for potential interest.

hvhaugwitz:
FWIW one can use repo-specific configuration to manage /etc files with vcsh:
```
$ mkdir -p .config/vcsh/config.d
$ echo "VCSH_BASE=/etc" > .config/vcsh/config.d/etc
$ vcsh init etc
$ vcsh etc add /etc/login.defs
```

```
$ vcsh version
vcsh 1.20151229
git version 2.31.0
```

PR #255, opened 2018-10-30
Set $XDG_CONFIG_HOME and $VCSH_BASE based on existing .config/vcsh

bschubernd:
Set $XDG_CONFIG_HOME and $VCSH_BASE based on existing .config/vcsh,
as discussed in vcs-home@lists.madduck.net 2018/10/28 in Mail with subject "dynamically set VCSH_BASE and XDG_CONFIG_HOME".

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -66,7 +66,7 @@ source_all() {
 # (To create a new $VCSH_BASE: mkdir -p $VCSH_BASE/.config/vcsh)
 if [ ! -n "$XDG_CONFIG_HOME" ] && [ ! -n "$VCSH_BASE" ]; then
   # Search for .config/vcsh in actual directory path
-  VCSH_BASE="$(realpath .)"
+  VCSH_BASE="$(pwd)"
   while :; do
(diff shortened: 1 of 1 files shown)


ao2:
Seems unnecessarily complicated, check out https://github.com/RichiH/vcsh/pull/115 which basically allows to set up `XDG_CONFIG_HOME` and `VCSH_BASE` as env variables in `./.config/vcsh/config` and enable sourcing that file from the current working directory.

Also related to https://github.com/RichiH/vcsh/issues/110

bschubernd:
> ble sourcing that file from the curr

You can continue to set XDG_CONFIG_HOME and VCSH_BASE as env variables in ./.config/vcsh/config.
#255 will change anything if you have set one of both variables.

But I do not want to set any variables in config files. This seems to me unnecessarily complicated. It just has to work automatically. The same way as it would work for the git command if a .git/ dir was found in the actual working directory path.

You can use "mkdir -p .config/vcsh" everywhere and start with new repositories. I have tested it with files in /srv/kvm/ and in /etc.

PR #259, opened 2019-03-26
Fixed debug function in Cygwin on Windows 10

arndtc:
The debug function was causing a bad exit code when calling vcsh from mr on Windows 10.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -139,7 +139,7 @@ help() {
 }
 
 debug() {
-	[ -n "$VCSH_DEBUG" ] && echo "$SELF: debug: $@"
+	if [ -n "$VCSH_DEBUG" ]; then echo "$SELF: debug: $@"; fi
 }
 
 verbose() {


alerque:
Thanks for taking the time to contribute this. Clearly the current code was wrong and I see you copied the solution from `verbose()`.

That being said I don't understand how it was causing a bad exit code. The `debug()` function itself could have return a failure in the event debugging wasn't enabled, but if it wasn't enabled then the code on line 16 should be returning a bad code too. The script isn't current set to run with the `-e` flag at all (a pet peve of mine, but a topic for another day). Is that mode somehow enabled in Windows by default? Or what am I missing here?

This fix itself seems right, but it might be an indicator of a larger problem we haven't found yet. Are you sure this fix was all that was required for it to work as expected

alerque (approved):

RichiH:
Thanks!

arndtc:
@alerque Thanks for looking.  I also wasn't able to track down why it was happening, but as you said, I copied it from the verbose() section and the fix solved the problem for me.  I've been using it for some time now, so thanks for accepting it.

PR #260, opened 2019-04-04
Recursively delete repository directories

soulofmischief:
This change allows vcsh to delete directories when removing a repository.

This fixes the situation where deleting a repository with submodules leaves behind submodules.

If a good case can be made for why `-r` should be optional, I can either rewrite delete to accept a `-r` option or introduce a separate recursive delete routine to vcsh.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -216,7 +216,7 @@ To continue, type 'Yes, do as I say'"
 	read answer
 	[ "x$answer" = 'xYes, do as I say' ] || exit 16
 	for file in $files; do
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for taking the time to contribute, but this runs way too high a rist of removing things people didn't intend to remove. I'm sure 99% of the time it wouldn't be an issue, but that 1% that stuffed something untracked in their submodule and blamed `vcsh` for nuking it is not worth the trouble.

If we were to go down this road this would need to much more cautiously first remove files then `rmdir` without any deploying any `rm -rf` nukes.

Thanks for understanding.

Issue #270, opened 2020-01-03
The zsh completion has a brain split


danielshahaf:
Compare:

https://github.com/RichiH/vcsh/commits/master/_vcsh
https://github.com/zsh-users/zsh/commits/master/Completion/Unix/Command/_vcsh

First of all, they are functionally different.  At least `vcsh foo add <TAB>` works in the former but not in the latter.

The latter has only one commit that isn't in the former: https://github.com/zsh-users/zsh/commit/13fc579343b24d298fb8905933b6000d7fcda114#diff-9394f5dc325d8a13da62fcaf4172a892  That commit unfortunately mixes functional and non-functional changes, but I think the important parts of it are two: adding `_call_program` and setting `ret`.  So I suggest to cherry-pick those two changes and then to remove `_vcsh` completion from zsh upstream.

danielshahaf:
Specifically, zsh commit https://github.com/zsh-users/zsh/commit/f9851d817fd6b45e00c35d71f9047f95eca06f42 has imported `_vcsh` as of commit 1d8970fdb8d4adbfd5cdecc5d15da9db61efb39d in this repository, verbatim:

```zsh
% diff -s =(cd vcsh && git cat-file blob 1d8970fdb8d4adbfd5cdecc5d15da9db61efb39d:_vcsh) =(cd zsh && git cat-file blob f9851d817fd6b45e00c35d71f9047f95eca06f42:Completion/Unix/Command/_vcsh) 
Files /tmp/zshwsSvMz and /tmp/zsh1po76w are identical
```

danielshahaf:
Created PR #271.  Once it's fixed and `_vcsh` is removed upstream, this issue may be closed.

alerque:
I've merged the related PR for this, but I'm not sure what to do about ZSH. Should we submit a PR to nuke the completions from their repository? Or on that brings it to parity with ours? It makes more sense to me as both a developer here and a distro packager to include it in our sources, but if upstream ZSH prefers to keep a copy themselves I guess the least we could do is update it. Is there a protocol they follow for this?

RichiH:
Historically, doing both was preferred, which is why I did it that way. Unclear if that's still true today, I stopped being active on the Zsh lists a decade ago.

@okapia, what's your opinion?

okapia:
Having it in both tends to just result in them diverging and it may not be ideal where projects have different licences. There have been a couple of other cases where we ended up removing the one in zsh.

[...]

PR #271, opened 2020-01-03
Completion: Fix the brain split (#270)

danielshahaf:
See #270 and the log messages for details.

@okapia FYI

### DIFF
--- a/_vcsh
+++ b/_vcsh
@@ -1,10 +1,9 @@
 #compdef vcsh
 
 function __vcsh_repositories () {
-	local expl
 	local -a repos
-	repos=( ${(f)"$(command vcsh list)"} )
-	_describe -t repos 'repositories' repos
+	repos=( ${(f)"$(_call_program repositories vcsh list)"} )
+	_describe -t repositories 'repository' repos
 }
 
 function __vcsh_not_implemented_yet () {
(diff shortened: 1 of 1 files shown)


alerque (commented):
Thanks for taking the time to contribute this @danielshahaf. Are you still using it yourself it this form? I'd like to test it myself before merging and am wondering if you have some test cases I could use each way to see what the effects are.

okapia:
@danielshahaf tagged me on this - I'm a zsh developer but not a vcsh user. From a read of the changed files I can fairly easily vouch for the fact that this looks clearly correct if that gives you some confidence in it. Not knowing vcsh, I can't really help on examples of differences. The change to return status makes it conform to a very common idiom where any function returning 0 should result in the function itself returning 0. This may need a non-trivial zsh configuration to demonstrate. If you use _approximate, you may get approximate completion happening when non-approximate completions were added by one called function but a subsequent one added nothing. This may only occur given suitably contrived matches so could be hard to reproduce but it defintely happens. The use of _call_program is good practice in general wherever completion functions call an external command to generate matches.

alerque:
Thanks for contributing to the review @okapia, those comments were helpful

alerque (approved):

danielshahaf:
Thanks, @okapia!

PR #273, opened 2020-04-17
A series of commits to improve bash completion

jamesd:
The first commit is the most important as it fixes a bug in how bash arrays are used.

The last commit improves completion of git subcommands by setting GIT_DIR appropriately.

The rest are cleanups (mostly from shellcheck).

### DIFF
--- a/_vcsh_bash
+++ b/_vcsh_bash
@@ -73,7 +73,7 @@ _vcsh () {
 	if [[ -z $cmd ]]; then
 		case $prev in
 			-c)
-				COMPREPLY=( $(compgen -f -- $cur) )
+				COMPREPLY=( $(compgen -f -- "$cur") )
 				return
 				;;
 		esac
@@ -81,11 +81,11 @@ _vcsh () {
 		case $cur in
 			-*)
 				OPTS='-c -d -h -v'
-				COMPREPLY=( $(compgen -W "${OPTS[*]}" -- $cur) )
+				COMPREPLY=( $(compgen -W "${OPTS[*]}" -- "$cur") )
 				return
 				;;
 		esac
-		COMPREPLY=( $(compgen -W "${repos[*]} ${cmds[*]}" -- $cur) )
+		COMPREPLY=( $(compgen -W "${repos[*]} ${cmds[*]}" -- "$cur") )
 		return 0
 	fi
 
@@ -96,7 +96,7 @@ _vcsh () {
 
 		list-untracked)
 			[[ $cur == -* ]] && \
-				COMPREPLY=( $(compgen -W '-a -r' -- $cur) ) && return
+				COMPREPLY=( $(compgen -W '-a -r' -- "$cur") ) && return
 			;;&
 
 		run)
@@ -109,7 +109,7 @@ _vcsh () {
 		delete|enter|list-tracked|list-untracked|rename|run|status|upgrade|write-gitignore)
 			# return repos
 			if [[ -z $subcmd ]]; then
-				COMPREPLY=( $(compgen -W "${repos[*]}" -- $cur) )
+				COMPREPLY=( $(compgen -W "${repos[*]}" -- "$cur") )
 				return
 			fi
 			return
@@ -117,13 +117,13 @@ _vcsh () {
 
 		clone)
 			[[ $cur == -* ]] && \
-				COMPREPLY=( $(compgen -W '-b' -- $cur) )
+				COMPREPLY=( $(compgen -W '-b' -- "$cur") )
 			return
 			;;
 
 		foreach)
 			[[ $cur == -* ]] \
-				&& COMPREPLY=( $(compgen -W "-g" -- $cur) ) && return
+				&& COMPREPLY=( $(compgen -W "-g" -- "$cur") ) && return
 			_vcsh_git_command $subcword
 			return
 			;;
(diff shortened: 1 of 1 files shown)


_vcsh_bash
```diff
@@ -128,11 +132,17 @@ _vcsh () {
...
-		_vcsh_git_command $subcword
+		local REPO_DIR="${XDG_CONFIG_HOME:-$HOME/.config}/vcsh/repo.d/${cmd}.git"
+		GIT_DIR=${REPO_DIR} _vcsh_git_command $subcword
 	fi
 	return 0
 }
 
 complete -F _vcsh vcsh
 
 # vim: ft=sh:
```
alerque, line 143: I think VIM modelines typically need to trail emacs ones, they need to be within a certain distance from the end.

_vcsh_bash
```diff
@@ -128,11 +132,17 @@ _vcsh () {
...
+# Local Variables:
+# mode:shell-script
+# sh-shell:bash
+# indent-tabs-mode: t
```
alerque, line 147: I'm okay with specifying the file type hints in a modeline, but lets leave the indentation hints for Editor Config. See #237 for where we just recently added that.

_vcsh_bash
```diff
@@ -128,11 +132,17 @@ _vcsh () {
 
 	# git command on repository
 	if [[ -n "${repos[$cmd]}" ]]; then
-		_vcsh_git_command $subcword
+		local REPO_DIR="${XDG_CONFIG_HOME:-$HOME/.config}/vcsh/repo.d/${cmd}.git"
```
alerque, line 135: Shouldn't this be using `$VCSH_REPO_D` or something similar instead of rolling your own guess as to the repository location?

_vcsh_bash
```diff
@@ -52,11 +52,15 @@ _vcsh_git_command () {
...
 	_init_completion -n = || return
 
-	local repos cmds
-	repos=( $(command vcsh list) )
+	local r reponames
+	local -A repos
+	mapfile -t reponames < <(command vcsh list)
+	for r in "${reponames[@]}"; do repos["$r"]="$r"; done
+	unset r reponames
+	local cmds
```
alerque, line 63: In `bash` can you declare a variable as local and define it in the same line? If so lets do that, but mostly I'm asking out of curiosity. On ZSH I'm accustomed to this being one operation.

_vcsh_bash
```diff
@@ -52,11 +52,15 @@ _vcsh_git_command () {
...
+	local cur prev words OPTS
 	_init_completion -n = || return
 
-	local repos cmds
-	repos=( $(command vcsh list) )
+	local r reponames
+	local -A repos
+	mapfile -t reponames < <(command vcsh list)
+	for r in "${reponames[@]}"; do repos["$r"]="$r"; done
+	unset r reponames
```
alerque, line 62: Just because I'm more into ZSH that Bash, any pointers on why these need explicit unsets?

alerque (changes requested):
Hey thanks for taking the time to contribute this. Most of it looks great. I have a few questions and nitpicks, plus one issue (the repo location variable issue) that might be significant.

I can probably take care of this stuff eventually with some `git revise` magic, but if you are around and still interested in this please feel free to jump in!

_vcsh_bash
jamesd: `unset` simply makes sure there is no name collision later in the function. It's a belt and suspenders thing really.

_vcsh_bash
jamesd: I'm just used to [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) and now declare and assign separately. In this case since it is a static list it is not required to be separate. I can make it a single line declare and assign if you like.

_vcsh_bash
jamesd: Ok, I will change the order.

_vcsh_bash
jamesd: Understood. Will remove that the tab setting.

_vcsh_bash
alerque: Thanks for the docs pointer here, that makes sense I guess. Carry on!

_vcsh_bash
alerque: Okay. It's moments like these that I wish shell code hygiene worked more like Rust, but there is no harm in explicitly doing your own housekeeping. Carry on.

_vcsh_bash
alerque: I didn't test it, but I'm pretty sure that's what the variable *is for*.

RichiH: Using `$VCSH_REPO_D` will be more robust than the proposed line.

jamesd: `VCSH_REPO_DIR` is [defined in _vcsh](https://github.com/RichiH/vcsh/blob/main/_vcsh#L101-L102) also since the `vcsh` script is not active for command completion. So really it is not much different than what I have other than the name of the local variable. I will rename it to be consistent with the other locations it is defined.

alerque: True, but it's defined in such a way that it's only set **if** it doesn't already exist. In the case of being inside `vcsh enter` it will be set in the host environment and that definition should be a NOOP. Using that variable is still the right think to do.

alerque (approved):
Thanks for polishing this up, it looks good to go to me.

RichiH:
Thanks!

PR #277, opened 2021-03-29
Switch testing from Travis to GitHub Actions

alerque:

PR #278, opened 2021-03-29
Switch testing from Travis to GitHub Actions

alerque:
This is the follow up to #277 after I embarrassing pushed to the wrong repo and hence killed off the PR.

This time the PR is coming from my fork and you can [see the tests in action here](https://github.com/alerque/vcsh/actions/workflows/test.yml?query=branch%3Agh-actions) even though the canonical repository won't run them until the workflow file exists in the default branch.


RichiH:
Tsk, killing cows...

Your GPG key needs a lifetime extension & push to keyservers, it seems.

RichiH (approved):

alerque:
Eh? Last I checked it wasn't 2022 yet. Then again the way the world is going I might have missed something. I'll look into why GitHub thinks I'm some unverified scammer.

alerque:
I just looked at half a dozen public key servers and my public key artifacts aren't expired on any of them. I don't know what GitHub is referencing, but I don't think it matters for this PR.

alerque:
@RichiH I finally figured out why GitHub was so grumpy about my GPG signatures. It wasn't that the key in my profile wasn't current –it was– but I had added a subkey since adding the public key to my profile. Their UI also has no support for updating public keys, but deleting and re-adding the public key convinced them the subkeys were all copacetic.

PR #281, opened 2021-04-01
Get debug information on remote CI's test runs

alerque:
Experimental follow up to test issues in #250

### DIFF
--- a/t/100-init.t
+++ b/t/100-init.t
@@ -24,7 +24,8 @@ $output = `./vcsh status`;
 
 ok $output eq "test1:\n\n", 'Our new repo is there';
 
-chdir $ENV{"HOME"} . '/.config/vcsh/repo.d/test1.git/' or die $!;
+chdir $ENV{"HOME"} or die $!;
+chdir '.config/vcsh/repo.d/test1.git/' or die $!;
 
 ok -f 'HEAD';
 ok -f 'config';
--- a/t/100-init.t
+++ b/t/100-init.t
@@ -18,7 +18,7 @@ ok $output eq "", 'No repos set up yet.';
 
 $output = `./vcsh init test1`;
 
-ok $output eq "Initialized empty Git repository in " . $ENV{'HOME'} . "/.config/vcsh/repo.d/test1.git/\n", "CROAK " . $output;
+ok $output eq "Initialized empty Git repository in " . $ENV{'HOME'} . "/.config/vcsh/repo.d/test1.git/\n", "CROAK " . $output . $ENV{'HOME'};
 
 $output = `./vcsh status`;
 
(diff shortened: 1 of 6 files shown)


alerque:
My Perl foo is to rusty for this.

Tests are failing in CI because either:

* `$ENV{'HOME'}` expands one way one second and a different one another second a few lines later in the same script with no apparent change to it; or
* `ok $output eq "a";` is **passing** when `$output = "b"`.

image

alerque:
Somebody please tell me the above wonkiness is not just Perl's idea of a bad April 1st joke.

RichiH:
Also, rusty but on a hunch: The first line is the croak of `$output` because the `ok` clause is not true.

t/100-init.t
```diff
@@ -18,13 +18,24 @@ ok $output eq "", 'No repos set up yet.';
 
 $output = `./vcsh init test1`;
 
-ok $output eq "Initialized empty Git repository in " . $ENV{'HOME'} . "/.config/vcsh/repo.d/test1.git/\n";
+ok $output eq "Initialized empty Git repository in " . $ENV{'HOME'} . "/.config/vcsh/repo.d/test1.git/\n", "CROAK " . $output;
```
RichiH, line 21:
```suggestion
ok $output eq "Initialized empty Git repository in " . $ENV{'HOME'} . "/.config/vcsh/repo.d/test1.git/\n", "CROAK " . $output . $ENV{'HOME'};
```

alerque:
Yup yup, for an hour at least messing around early I missed the fact that that `ok` was actually failing. I was so hung up on the later failure that gave a related error message (directory not existing) I missed the fact that wasn't just a success echo it was an error.

As for your test, that shows Git's understanding of `$HOME` isn't comming from Perl:

> Failed test 'CROAK Initialized empty Git repository in /home/runner/.config/vcsh/repo.d/test1.git/ /home/runner/work/vcsh/vcsh/t/etc/.vcsh_home'

On my local system the `$ENV{'HOME'}` value is respected. The test repos are showing up in the right place not in my user's home dir.

RichiH:
The reason why I chose Perl's tests back then is because nothing else did CLI testing nicely. Is there anything new and easier to hold we could just replace the whole thing with?

alerque:
Yes, quite a few options actually. And some work has gone into setting one up for this, I just haven't gotten there yet. It just rubs me the wrong way that I can't figure out this one little oddity.

RichiH:
Ooohhhh!

`$ENV{'HOME'} = abs_path ('.vcsh_home');`

What if the new test system prevents us from changing `$HOME` as a defense mechanism?

RichiH:
https://github.com/RichiH/vcsh/pull/281/checks?check_run_id=2249701727#step:5:39 tells me

1. that changing `$HOME` is allowed
2. that `abs_path` is broken, somehow

alerque:
Eureka! I think I've got this, stand by for fireworks.

alerque:
Some days the environment breaks you. But on the third day's morning coffee you break the environment.

RichiH:
Heh; nice find!

(In the general case, please give me some time to review after flagging as ready, but in this case, it's self-evidently correct)

PR #282, opened 2021-04-02
Make clone() check GIT_REMOTE early

### LINKED ISSUE #161: When ssl validation fails vcsh assumes remote is empty
Requirement to recreate: No ca-certificates are installed:
    # vcsh clone https://github.com/RichiH/zshrc.git
    Initialized empty shared Git repository in /root/.config/vcsh/repo.d/zshrc.git/
    Switched to a new branch 'master'
    vcsh: info: remote is empty, not merging anything.
      You should add files to your new repository.

Seems like successful clone (of an empty repo), but
    # vcsh pull
    zshrc: fatal: unable to access 'https://github.com/RichiH/zshrc.git/': Problem with the SSL CA cert      (path? access rights?)

shows the error. 
IMHO VCS should throw that error on initial clone.

### LINKED ISSUE #209: Fail cloning when the remote repository does not exist
While cloning, before we initialize the local repo we should first check whether the remote repository exists or not. I made a mistake while typing the local repo path when cloning and the cloning proceeded creating a repository with empty contents, not failing at any step. It only failed on a subsequent pull command. This can be avoided if we check right in the beginning using `git ls-remote -h <path>` and fail the clone if the repo does not exist.

alerque:
Closes #161
Closes #209

This very sensible proposed fix to a long standing issue never made it into the default branch and hence not into releases. I can't see any good reason not to do this, the slight overhead has to be worth it to avoid dirty repository states. Clone isn't a frequent enough operation to make the compromise.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -149,6 +149,8 @@ info() {
 
 clone() {
 	hook pre-clone
+	# Check if remote is reachable. Abort early if there's a typo, TLS certificate problem, etc
+	git ls-remote "$GIT_REMOTE" 2> /dev/null || fatal "Can not reach '$GIT_REMOTE'"
 	init
 	git remote add origin "$GIT_REMOTE"
 	git checkout -b "$VCSH_BRANCH" || return $?


RichiH:
I thought I had merged that half a decade ago.. Thanks!

Issue #166, opened 2015-05-31
please store binary blobs in $XDG_DATA_HOME, not $XDG_CONFIG_HOME


bremner:
For me, config is something I can edit with a text editor. So I think the repos belong on $XDG_DATA_HOME.

I admit that 1) this is pretty subjective. 2) most software that uses XDG_CONFIG_HOME does it wrong according do my definitions. But we can do better, right ? ;)

RichiH:
As per IRC, the main problem is to deal with legacy. My only real answer is
a softlink...

Richard

Sent by mobile; excuse my brevity.

bremner:
Richard Hartmann <notifications@github.com> writes:

> As per IRC, the main problem is to deal with legacy. My only real
> answer is a softlink...

Let me see if I understand the problem.  The problem is changing the
default for VCSH_REPO_D, and having new vcsh unable to find old repo.d ?
what about having it look in both places if no explicit value is given
for VCSH_REPO_D ?  I guess it's not so nice to special case that
variable; maybe you could convert it to a path, then give a default
value with $XDG_DATA_HOME first?

d

RichiH:
What happens when, not if, someone uses an old release with a migrated
$HOME?

I only see symlinks as a way out.

Richard

Sent by mobile; excuse my brevity.

bremner:
Richard Hartmann <notifications@github.com> writes:

> What happens when, not if, someone uses an old release with a migrated
> $HOME?
>
> I only see symlinks as a way out.
>

They lose. I would find that acceptable, but I guess you don't.

d

nfarrar:
I would agree that $XDG_CONFIG_HOME is not the proper place for the repo.d files. However - I also do not think that $XDG_DATA_HOME is correct.

While these are 'binary blobs' - they aren't data you'd want to backup or transport to another system. I believe it would be *most correct* to stored them in `$XDG_CACHE_HOME`.

4nd3r:
> $XDG_CACHE_HOME defines the base directory relative to which user specific **non-essential** data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

not really sure if repository is non-essential. things shouldn't break if you delete `.cache` content.

RichiH:
cache is not an option; it's ephemeral by definition...

@bremner I think I can use the new path for new clones etc, then check both paths every time I do any action and print a warning if both exist. Sounds sane?

bremner:
[...]

PR #284, opened 2021-04-02
Make path to .gitignore.d and .gitattributes.d configurable

RichiH:
Reviving https://github.com/RichiH/vcsh/pull/203 with some dark magic.


alerque (changes requested):
Mostly I think this code looks like it is on the right track as far at it goes. Before we merge it however I think we should at least add a check that throws a warning if the default settings are changed but the legacy default location still has matching files, then perhaps extend `vcsh upgrade` with a fix.

In fact I might even make a case that in addition to providing this configurable option we should change the default to be somewhere in `$XDG_DATA_HOME`, cf. #166.

PR #286, opened 2021-04-03
Fix broken getopts() usage, allows multiple flags

alerque:
The current usage of getopts is a kind of dirty hack: using it to iterate over passable arguments but using a hand rolled flag matching syntax. It appears this was done as a workaround to catch a long form ‘--debug’ argument, but I believe it introduces more problems than it solves. The long form argument does sort of work, but only after getpots throws an error about invalid argument syntax and only on a fluke of ‘-d’ being a substring of ‘--debug’ and getpots having assigned ‘ebug’ as $OPTARG.

In one sense this is a regression because it removes the long form option parsing, but it was not documented in help anyway and I believe it is unlikely anybody is using this particular flag in a script that will break.

On the other hand with proper iteration of the getopts assigned flags variable, we can now properly handle multiple flags. Previously things like `-d -v` or `-v -c foo` would attempt to do crazy broken things such as:

    GIT_DIR=/home/caleb/.config/vcsh/repo.d/-v.git

Now multiple flag handling just works as expected.

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -34,21 +34,25 @@ fatal() {
 # We need to run getops as soon as possible so we catch -d and other
 # options that will modify our behaviour.
 # Commands are handled at the end of this script.
-while getopts "c:dv" flag; do
-	if [ x"$1" = x'-d' ] || [ x"$1" = x'--debug' ]; then
-		set -vx
-		VCSH_DEBUG=1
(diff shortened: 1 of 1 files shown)


RichiH (approved):
Nice!

Issue #202, opened 2016-05-13
One-liner bootstrap


jhgorse:
Pursuant to a ssh-copy-id like solution for setting up a new account.
## My bootstrap:

wget https://raw.githubusercontent.com/RichiH/vcsh/master/vcsh
chmod +x vcsh
./vcsh clone https://github.com/jhgorse/bash.git bash # Perhaps a recursive clone some day
./vcsh run bash submodule init
./vcsh run bash submodule update
## Proposed:

sh "$(curl -fsSL https://raw.githubusercontent.com/RichiH/vcsh/master/vcsh)" clone --recursive https://github.com/jhgorse/bash.git bash

or perhaps a separate boostrap script that adds a ~/bin or ~/.bin directory, vcsh, chmod+x, then calls the trailing bootstrap args upon setup completion.

RichiH:
Personally, I don't like to cURL into a shell, but if you want your own personal bootstrap, wouldn't it make sense to just have a shell script you can cURL?

`curl https://my.dom/script.sh | sh`

or the like.

jhgorse:
@RichiH what is your preferred manner of bootstrapping a new environment?

alerque:
I think I'm going to chime in strong agreement with Richard on this one. I don't think public projects should endorse the `sh <(curl <something>)"` paradigm. It's just so dangerous on so many levels... If you're going to do that kind of thing you should be responsible for both ends of the stick. Ironically [my own bootstrap method](https://github.com/alerque/que/blob/master/bin/que-home-bootstrap.bash) involves something like this, but at least I am responsible for my own destiny. If I make a bum turn...

![image](https://user-images.githubusercontent.com/173595/112896710-6322c400-90e7-11eb-8be1-9d40a7559c39.png)

Keep in mind that there are two parts here: you need a system with `vcsh` itself, and you need your own personal dotfiles repository. We don't want to meddle much with the latter, and the former is highly system dependent. Getting `vcsh` on a system should be done through that system's package manger. If any bootstrapping is done it should really be done tailored to your system, so again I think hosting your own preferred jump starter kit is probably better than anything this project can provide upstream.

If you have ideas for how this could be improved without compromising ⓐ security best practices and ⓑ flexibility please do comment and we'll reconsider.

jhgorse:
So what then is the purpose of vcsh if it is not to improve the bootstrapping experience of a new shell?

[...]

PR #288, opened 2021-04-03
Setup autotools to configure and build

### LINKED ISSUE #41: TEST: Create test that tagged version is returned by `vcsh version`

alerque:
Closes #167
Closes #41

This is a stab at converting the custom makefile to autotools. Dependencies are checked via autotools and the system discovery bits it comes up with are used throughout.

~~I'll say a bit more about this and update relevant docs, but I am at a stopping point for the night and need to check CI results next anyway.~~

**Edit:* See pros & cons below the rule, and comments following [here](https://github.com/RichiH/vcsh/pull/288#issuecomment-817211254).

----

### Downsides

- Installing from Git clones requires an extra step, and developers plus anyone wishing to install from a git clone need to have GNU Autotools. The usual route to just download and install will be from a source package instead. The extra step to build from Git clones is to run `./bootstrap.sh`. For `git archive` generated source snapshots which have neither a pre-packaged version embedded in the source nor Git history from which to guess there are *two* extra steps: `echo YOUR_VERSION > .version; ./bootstrap.sh`.

- The process of cutting a release is ... different. I'm not sure this is a downside because it is easier to automate than before and involves no special handling for any distros — but it will require different steps. Listing this just because any change in procedure can be a downside at least for the transition process...

- The process of packaging VCSH will be different, requiring all existing packagers to adjust. As a distro packager myself I think this will me mostly welcome, but it will slow the speed of propagation. Any hacks to fix install locations, not generate man pages, skip tests, etc. can all be removed in favor of a *very* generic `./configure && make install`.

- The final installed script is *less* portable than it was. Because it will be custom tailored to the system on which it was initially installed, it can't be universally copied to any other system and expected to work as well. It is still a self contained script and can be copied to similar systems, but coping it across different OSs may or may not work depending on how similar they are. A script from a system with `/usr/bin/git` with  This has a negative impact on #202, but if a bootstrap variant is needed we can still generate one with everything relying on `$PATH` based instead of compile time discovered paths. This isn't hard (`./configure SED=sed GIT=git ...`) the real question will be where to *put* such a portable variant.

- The source input script is a template, not itself the final script. This adds a little friction during the development cycle because you can't run it for quick tests without running `make` to update the output form. Additionally some tooling won't work quite as automatically, e.g. `shellcheck` needs to be run on the final form of the script instead of the input template and editor tooling like ALE won't run it automatically. This is mitigated by the fact that it runs in CI (so it won't be permanently forgotten) and quick hacks can be tested by editing the output script, then porting whatever works to the template, but it's not *quite* as simple as before.

### Features

- During the release process (and in between) there is no need to edit anything to get the correct VERSION string compiled into the script. No more forgetting to bump the version or installations that self identify as some version older than they are.

- There is no need for anybody but developers to have `ronn` or anything else Ruby. The `make dist` process will generate a source tarball with a ready-made man page. Those wishing to install using a Git clone instead of a source tarball can use `./configure --without-man-page` to skip the dependency checks and relevant build bits.

- There is no need for anybody but developers to have `perl` or and of the related test modules. Source tarballs won't run tests by default (although they can be configured to with `./configure --enable-tests`), only a light self check. Developers using a Git clone will get test dependencies and targets by default but can disable them with `./configure --disable-tests`.

- Broader target system compatibility: autotools generated makefiles are *extremely* robust, easily the most robust single way to target POSIX systems. Many differences the current manual `Makefile` won't handle are taken care of automatically.

- Autocompletions are installed to paths that systems can configure using `pkg-config` or than can be overridden with a configure flag. They can also be disabled completely.

- Partial installs of, for example, just the binary can be done with `make install-exec`.

- Deals with every ‘sin’ on the dirty laundry list listed in #167, including ones that the manual Makefile changes wouldn't have addressed and without GNU vs. BSD tooling issues.

- The final name of the program can be prefixed, suffixed, or renamed again using standard tooling. This can be useful to install multiple versions, to install a user local version that doesn't conflict with a system version, etc. Uses the standard `./configure` args `--program-prefix`, `--program-suffix`, or `--program-transform-name`.

- Any or all dependencies may be supplied manually rather than detected at configure time. For example, you may have an older version of Git installed by default on your system, but a new version available somewhere else. Using `./configure GIT=/path/to/git` before `make install` is enough to specify a specific path that `vcsh` will use whenever it invokes Git.


RichiH:
What about `rm` and `mkdir`? Should they be relied upon to be in `$PATH` or also found through autotools?

alerque:
> What about `rm` and `mkdir`? Should they be relied upon to be in `$PATH` or also found through autotools?

Good question, and I was actually researching that this morning. I started second guessing it after the work I did the other day.

I'm pretty sure the answer is no, and the logic is that a POSIX environment is assumed and certain coreutils can be assumed to be a part of `$PATH`. The issue with `sed`→`$(SED)`, `awk`->`$(AWK)`, `make`→`$(MAKE)` and a few others is that many systems have more than one implementation lying around and `autoconf` does a bunch of work to make sure the most appropriate ones are being used even if that is not the first match in `$PATH`.

This matters particularly on platforms like macOS that have half BSD tooling and half old GNU tooling (pre GPL3). Plain macOS and Homebrew end up with different assortments of linked tools and path priorities. Using the autoconf variables matters most when it's possible to have different implementations (e.g. `sed` argument handling carries between BSD/GNU variants). The same does not not hold in the case of `rm` and `mkdir`.

For `git`→`$(GIT)` and other non-core utilities the reason I implemented them is it makes sure the installed component uses the same thing that the compile time dependency check found. That way we can add a minimum version check and/or rely on what we discovered about it. Again on macOS and others the system default is often not what Homebrew teaches autoconf to find.

All this gives end users (and distros) a standardized way to adjust all these values. Homebrew is a good example because it teaches autoconf where to find its internal bits, but that goes out the window if you run the script later outside of its protection and this time get a different `git` version. Even if your average Joe (excluding Joey of course) isn't going to know how to substitute a different Git at compile time, the `./configure` options work the same as for any other autoconf based system and many distros have this all plumbed up by default.

alerque:
Just FYI @RichiH I did some fancy merging from `release`→`main`. The merge brings the commit and tag history but I modified the merge itself to not bring in the release artifacts such as the man page and Makefile changes. I had to do this directly rather then as a GitHub PR because it won't let you resolve merge conflicts it doesn't think exist.

This step was necessary for a couple of reasons:

* The changelog was out of date on `main`, and merging the complete one seemed saner than trying to sync them.
* The git history wasn't reflecting the tag sequence. Notably `git describe` was coming up with something that didn't have the latest safe harbor tags because they were tagged on an unrelated branch. AS of *this* PR tag history becomes the only way the script knows what version it is. Now for example `git describe --always --tags` comes up with `v1.20190621-74-g4cc623f` — which identifies the last tag, the 74 commits not found in the tag, and the actual SHA which is just what `./configure` is going to expect to find.

Just mentioning it for the record since it was a direct merge/push.

RichiH:
Aye, I saw. I hate rebasing, but wondered if it would make sense in this case.

alerque:
I *love* rebasing and do it all day long (also see `git revise` which is fantastic to make the process faster) but I don't think it ever makes sense to rebase either the default branch or anything with a release tag in it—ever. As such the tags, the release branch, and main must all be attended to without rebasing.

Feature branches on the other hand I do all the time. This one has been rebased many times so far (including across the merge boundaries) and may be so again before it's ready.

alerque:
Okay I think I'm ready for real reviews on this. I've fed this through a few meat grinders myself but there is nothing like an outside pair of eyes for bringing up potential problems.

Note I've edited the [initial PR post](https://github.com/RichiH/vcsh/pull/288#issue-608456503) with a list of pros and cons I could thingk off off the top of my head after working on this.

I can rebase this to reduce the commit count a bit if desired, about half the later commits could be fixups for earlier ones.

alerque:
@leycec definitely open to other reviews here since clearly you've had more than a cursory experience with `make` before ;-)

alerque:
Just a note, `doc/INSTALL.md` is updated, but the TL:DR; for developers is that if you are using a Git clone, run `./boostrtap.sh` at least once after cloning or of course after you get agressive with `git clean -dxff` or anything like that. This rebuilds the autotools scripts that are not specific to this project. These are included in the source packages generated by `make dist` but not tracked in the this repository.

alerque:
Hey @dastergon many moons ago you had a PR #100 that got subsumed by #167 and now by this PR. Yes it's been a long road to get the configure and build steps fixed but I'm hoping we're near the end of it. Anyhow your concern was with the ZSH completions install directory on ... whatever distro you use. Any chance you could test drive this PR to see if it puts things in the right place for you? You should be able to do that without actually installing anything. Instead of `make install` and looking for files you can just exfiltrating the discovered location after configure like this:

```console
$ gh checkout pr 288
$ ./bootstrap.sh
$ ./configure --prefix=/usr
$ make -f Makefile -f <(echo 'debug:\n\t@echo $(zshcompletiondir)') debug
```

alerque:
@RichiH Without asking you to actually *like* autotools (or anything that smells like `auto*#!?`) can I ask how this is sitting with you? Particularly the pros/cons listed in the opening comment and the current state of the build process in this PR ... assuming you've had a chance to look at it.

alerque:
Ping @RichiH, see :arrow_up:.

alerque:
Hearing no objections....

I the interest of being able to work on other aspects without repeatedly breaking this PR, being able review and move on with other PRs, and hopefully get a few more eyeballs on this before the next release I'm going to go ahead and merge. That will loop in people that are using Git HEAD builds instead of tagged releases.

nferch:
Is this the right place to discuss the pros and cons of this change given that it has already been merged?

In short, the portability of the script and the lack of a build were one of the most attractive aspects of this project. With just curl/wget and git I could bootstrap my shell environment regardless of the flavor and state of the system within a few seconds. It's desirable to avoid any additional steps or dependencies since in many cases, the reason I need to log into the system at all is to address issues with the system that could prevent vcsh building or bootstrapping.

alerque:
Hey @nferch. I think the right place for that discussion is probably now issue #202. I'll also just note that the `vcsh` script itself is still completely portable between any *like* systems (i.e. ones that have executables in the same paths). Until we post a single line bootstrap version somewhere the biggest difference is that in order to bootstrap a system you should probably just `scp` a copy from one of your machines since you can't pull it straight out of this repository.

That being said I'm quite surprised to hear you have a use case for getting `vcsh` working as a way to recover a system that has broken Autoconf tooling. How exactly could being able to bootstrap this script get you out of a bind with broken core Unix system tooling?

nferch:
Thanks @alerque I'll comment in https://github.com/RichiH/vcsh/issues/202

PR #289, opened 2021-04-04
Update mailmap

RichiH:
Signed-off-by: Richard Hartmann <richih@richih.org>


alerque (approved):
Cool beans. I'll surely rebase my branch in #288 again before it's ready so I'll refresh the contributors list commit in there to catch this.

Also @lyda could we trouble you for a line for yourself? We've got you showing up under two emails in the contributors list as well. If there is one you'd like to see as canonical there we can use this map to keep that list clean.

PR #290, opened 2021-04-04
Pick an email address Lyda

lyda:
Since I'm a lazy typist, I have picked the shortest one.


alerque (changes requested):
`CONTRIBUTORS` is auto-generated, but regenerating it right now adds *two* entries for you. Sorry if it wasn't clear from #289 but what I was looking for was a new entry in `.mailmap` that consolidates them so the next time we regenerate it you only have one instance (unless you have a clone we should know about).

lyda:
Fixed.

alerque (approved):

alerque:
ty. ← *(me trying to conserve keystrokes)*

PR #292, opened 2021-04-05
Make VCSH_* availaible to all hooks

RichiH:


RichiH:
NB: I changed the version number to 0 (so it's always below our current 1.) 2021 (so it's a rough year) GIT (so it's clear where it's from). This might be made useless by the autotools PR when/if merged, but seems like a good failsafe for now.

alerque (approved):

Issue #287, opened 2021-04-03
Fix option parsing so it works across possible SH implementations


alerque:
Per the yucky discovery [here](https://github.com/RichiH/vcsh/pull/285#issuecomment-812884909) our current option parsing is *not* robust across SH implementations. Whatever @RichiH's has works. My systems don't.

On my systems `/bin/sh` is a symlink to bash, which should be triggering its internal POSIX compatibility mode. Whatever you think of this it's a pretty common arrangement.

On the other hand I tested `dash` and I get different results:

```console
$ ./vcsh -v foreach -g echo foo
verbose mode on
vcsh 1.20141026
vcsh: verbose: foreach begin
que-atom:
git: 'echo' is not a git command. See 'git --help'.

The most similar command is
        fetch

$ dash ./vcsh -v foreach -g echo foo
verbose mode on
vcsh 1.20141026
vcsh: verbose: foreach begin
que-atom:
foo
```

RichiH:
I seem to remember very, very, very dimly that Bash's POSIX compatibility mode not actually being POSIX compatible is what prompted me to hand-craft the option parsing.

@mirabilos might be able to help with his POSIX-fu.

alerque:
It's possible that is the case, but the hand crafted version isn't working either. Different error (it fails early) but it still doesn't work. The safe-harbor release under Bash's pseudo POSIX gives:

```console
$ vcsh -v foreach -g echo foo
verbose mode on
vcsh 1.20141026
/usr/bin/vcsh: illegal option -- g
...
```

PR #293, opened 2021-04-05
Shellcheck appeasement

### LINKED ISSUE #249: Run shellcheck and checkbashisms
Hi,

I noticed some messages when running `shellcheck` or `checkbashisms` on `vcsh`, there are some false-positives but there are also some useful suggestions.

Would you accept patches to address those?

Ciao,
   Antonio

### LINKED ISSUE #279: Refactor to please shellcheck / shellharden
Split out of #249...

I have mixed feelings about `shellcheck`/`shellharden`. I don't always like the style choice mandated even in cases where a live person can verify that no problem is present, but I do like the normalization and consistency. They can actually save a lot of grief chasing elusive bugs and security issues.

In any event, **if** we do go down that road, [as I mentioned](https://github.com/RichiH/vcsh/issues/249#issuecomment-810066745) I'd prefer to do it sometime after the PR churn has settled down a bit. We have quite a few things to consider before doing something that would make merges that much more work.

lyda:
Closes #249
Closes #279

This results in a clean shellcheck pass. Not sure how to then add it to the workflow...

### DIFF
--- a/vcsh
+++ b/vcsh
@@ -544,7 +544,7 @@ write_gitignore() {
 	IFS=$(printf '\n\t')
 	gitignores=$(for file in $(git ls-files); do
 		if [ "$GIT_VERSION_MAJOR" -ge 2 ] \
-		   && [ "$GIT_VERSION_MINOR" -ge 7 ]; then
+				&& [ "$GIT_VERSION_MINOR" -ge 7 ]; then
 			echo "$file";
 		else
 			while true; do
--- a/vcsh
+++ b/vcsh
@@ -543,7 +543,7 @@ write_gitignore() {
 	use
 	cd "$VCSH_BASE" || fatal "could not enter '$VCSH_BASE'" 11
 	# Works in all shells we care about.
-	# shellcheck disable=SC2039
+	# shellcheck disable=SC2039,SC3043
 	local GIT_VERSION GIT_VERSION_MAJOR GIT_VERSION_MINOR
 	GIT_VERSION="$(git --version)"
 	GIT_VERSION_MAJOR="$(echo "$GIT_VERSION" | sed -E -n 's/.* ([0-9]+)\..*/\1/p')"
--- a/vcsh
+++ b/vcsh
@@ -179,9 +179,7 @@ clone() {
 	git checkout -b "$VCSH_BRANCH" || return $?
 	git config branch."$VCSH_BRANCH".remote origin
 	git config branch."$VCSH_BRANCH".merge  refs/heads/"$VCSH_BRANCH"
-	# BSD wc emits spaces.
-	# shellcheck disable=SC2046
-	if [ $(git ls-remote origin "$VCSH_BRANCH" 2> /dev/null | wc -l ) -lt 1 ]; then
+	if [ "$(git ls-remote origin "$VCSH_BRANCH" 2> /dev/null | wc -l )" -lt 1 ]; then
 		info "remote is empty, not merging anything.
   You should add files to your new repository." # editorconfig-checker-disable-line
 		exit
--- a/vcsh
+++ b/vcsh
@@ -20,15 +20,15 @@
 # version of vcsh; the master branch is supposed to be clean at all times
 # so you can most likely just use it nonetheless
 VERSION='1.20141026'
-SELF=$(basename $0)
+SELF="$(basename "$0")"
 
 # Ensure all files created are accessible only to the current user.
 umask 0077
 
 fatal() {
 	echo "$SELF: fatal: $1" >&2
-	[ -z $2 ] && exit 1
-	exit $2
+	[ -z "$2" ] && exit 1
+	exit "$2"
 }
 
 # We need to run getops as soon as possible so we catch -d and other
@@ -56,7 +56,8 @@ shift $((OPTIND-1))
 
 source_all() {
 	# Source file even if it's in $PWD and does not have any slashes in it
-	case $1 in
+	# shellcheck source=/dev/null
+	case "$1" in
 		*/*) . "$1";;
 		*)   . "$PWD/$1";;
 	esac;
@@ -65,10 +66,12 @@ source_all() {
 
 # Read configuration and set defaults if anything's not set
 [ -n "$VCSH_DEBUG" ]                  && set -vx
-: ${XDG_CONFIG_HOME:="$HOME/.config"}
+: "${XDG_CONFIG_HOME:="$HOME/.config"}"
 
 # Read configuration files if there are any
+# shellcheck source=/dev/null
 [ -r "/etc/vcsh/config" ]             && . "/etc/vcsh/config"
+# shellcheck source=/dev/null
 [ -r "$XDG_CONFIG_HOME/vcsh/config" ] && . "$XDG_CONFIG_HOME/vcsh/config"
 if [ -n "$VCSH_OPTION_CONFIG" ]; then
 	# Source $VCSH_OPTION_CONFIG if it can be read and is in $PWD of $PATH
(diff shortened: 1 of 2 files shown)


vcsh
```diff
@@ -20,15 +20,15 @@
 # version of vcsh; the master branch is supposed to be clean at all times
 # so you can most likely just use it nonetheless
 VERSION='1.20141026'
-SELF=$(basename $0)
+SELF="$(basename "$0")"
```
RichiH, line 23: This quoting seems off.

vcsh
```diff
@@ -56,7 +56,8 @@ shift $((OPTIND-1))
 
 source_all() {
 	# Source file even if it's in $PWD and does not have any slashes in it
-	case $1 in
+	# shellcheck source=/dev/null
```
RichiH, line 59: What does this mean?

vcsh
```diff
@@ -65,10 +66,12 @@ source_all() {
 
 # Read configuration and set defaults if anything's not set
 [ -n "$VCSH_DEBUG" ]                  && set -vx
-: ${XDG_CONFIG_HOME:="$HOME/.config"}
+: "${XDG_CONFIG_HOME:="$HOME/.config"}"
```
RichiH, line 69: This seems to leave `$HOME/.config` unquoted

vcsh
```diff
@@ -146,11 +149,13 @@ help() {
 # editorconfig-checker-enable
 
 debug() {
-	if [ -n "$VCSH_DEBUG" ]; then echo "$SELF: debug: $@"; fi
+	if [ -n "$VCSH_DEBUG" ]; then echo "$SELF: debug: $*"; fi
```
RichiH, line 152: I will need to read up on `$@` vs `$*` again

RichiH (commented):
Quite a few of those seem very useful; but a few seem to unquote the dangerous parts completely.

Somewhat torn on the tool-specific comment; is it at least possible to put them all on one line?

vcsh
RichiH: Ah, wait, maybe it does not.

vcsh
lyda: Nope, for the shell expressions `${ }` and `$( )` it understands interior quotes. This can be confusing since it does not do this elsewhere (looking at you backticks) but the parens and curly braces give the parser more to work with.

vcsh
lyda: To fully parse things, shellcheck needs to follow `. some.file.sh` lines. If the file doesn't exist, you can point it at one. I cheat here and say "ignore this" by feeding it `/dev/null`.

vcsh
lyda: `$@` is an array, `$*` is a string. And generally you only want to use them in double quotes. In fact, I can't think of a single use case for an unquoted `$@`.

```
cat foo.sh             
#!/bin/bash

for i in "$*"; do
  echo "\"\$\*\": $i"
done
for i in $*; do
  echo "\$\*: $i"
done
for i in "$@"; do
  echo "\"\$@\": $i"
done
for i in $@; do
  echo "\$@: $i"
done
$ ./foo.sh 1 2 3 "4  5" 6
"$\*": 1 2 3 4  5 6
$\*: 1
$\*: 2
$\*: 3
$\*: 4
$\*: 5
$\*: 6
"$@": 1
"$@": 2
"$@": 3
"$@": 4  5
"$@": 6
$@: 1
$@: 2
$@: 3
$@: 4
$@: 5
$@: 6
```

vcsh
alerque: I agree it *seems* off, but have also been down the rabbit trail and know it to be correct. It would be pretty vanishingly rare to have a space in the name, but this is technically correct. I'm fine with being a bit pedantic to appease the tooling.

lyda:
> Somewhat torn on the tool-specific comment; is it at least possible to put them all on one line?

Possibly. My version of shellcheck didn't support it but supposedly [it can] so I might be out of date.  Just upgraded my desktop from 18.04 to 20.04 so will see if it likes single line comments now.

As an aside, I use [ALE] which is quite nifty. Very much recommend it. Linters run all the time no matter what language I'm coding in and get flagged in vim as I go.

[it can]: https://github.com/koalaman/shellcheck/wiki/Directive
[ALE]: https://github.com/dense-analysis/ale

vcsh
alerque: Same as [here](https://github.com/RichiH/vcsh/pull/293#discussion_r607069450).

vcsh
alerque: POSIX sh doesn't have arrays. Your examples here are Bash based.

lyda: POSIX shell doesn't have arrays but it does draw a distinction between `$*` and `$@`. It's where `bash` got its array syntax from. My example would work the same in dash (`/bin/sh` on Ubuntu) or on a posix shell on a BSD or vendoer Unix system.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_05_02

alerque: My bad, my comment is only half right. POSIX doesn't specify have arrays, it specifies array—singular. `$@` is the only available one.

vcsh
```diff
@@ -170,6 +175,8 @@ clone() {
 	git checkout -b "$VCSH_BRANCH" || return $?
 	git config branch."$VCSH_BRANCH".remote origin
 	git config branch."$VCSH_BRANCH".merge  refs/heads/"$VCSH_BRANCH"
+	# BSD wc emits spaces.
```
alerque, line 178: Yes. So? I don't think quoting this hurts, `[ " 2 " -eq 2 ]` tests the number just fine, we can probably go with it's recomendation and not worry about the override.

alerque (changes requested):
I think we'll need an override for [SC2220](https://www.shellcheck.net/wiki/SC2220) as well, we deliberately *don't* want to catch invalid flags. At least until I figure out what to do about #287.

alerque:
I'd also prefer linewise overrides where possible. Requiring the latest version of `shellcheck` is fine since its a developer only thing and we will run it in CI anyway.

RichiH:
Pulling that test to top level:

````
richih@beast  /tmp % which /bin/sh
/bin/sh
richih@beast  /tmp % cat fooz.sh
#!/bin/sh

for i in "$*"; do
  echo '"$*:" ' "$i"
done
for i in $*; do
  echo '$*: ' "$i"
done
for i in "$@"; do
  echo '"$@:" ' "$i"
done
for i in $@; do
  echo '$@: ' "$i"
done
richih@beast  /tmp % ./fooz.sh 1 2 3 "4  5" 6
"$*:"  1 2 3 4  5 6
$*:  1
$*:  2
$*:  3
$*:  4
$*:  5
$*:  6
"$@:"  1
"$@:"  2
"$@:"  3
"$@:"  4  5
"$@:"  6
$@:  1
$@:  2
$@:  3
$@:  4
$@:  5
$@:  6
richih@beast  /tmp % 
````

vcsh
RichiH: https://github.com/RichiH/vcsh/pull/293/commits/b66ed4e723ff58f4ad39477fd3257caff5ba7fe0 behaves correctly.

RichiH: https://github.com/RichiH/vcsh/pull/293/commits/b66ed4e723ff58f4ad39477fd3257caff5ba7fe0 behaves correctly and is not nagged at by shellcheck.

RichiH:
I enabled reviewdog for this repo and then got: 
image

I asked GH support for clarification what, exactly, the scope of this would be.

alerque:
It has to do with repo scope and whether tokens are allowed to run on actions coming from forks. The automatically supplied runner ones are not these days. I'm trying to find the last place I successfully set ReviewDog stuff up. I think I used a PAT of my own rather than the automatic one but I think they might have blocked that too.

We can run `shellcheck` on Ubuntu directly in the workflow its just more fun to get PR review comments with docs link than red marks and logs.

alerque:
@RichiH I don't have access to add secrets to the Actions for this repo. I think you need to [generate one](https://github.com/settings/tokens) then [add it here](https://github.com/RichiH/vcsh/settings/secrets/actions). If you use the name `REVIEWDOG_PAT` I think the workflow I just setup will start working. Maybe. Fingers crossed.

RichiH:
No dice :(

image

RichiH:
Giving up for today:
image

RichiH:
Side note, this PR will need to squash merged; it's a complete mess by now :p

lyda:
Why hide the fun?

alerque:
1. You don't need commits to trigger test runs, you can just hit "rerun" from the checks tab of this PR.

2. Git history is mutable, I've already cleaned this up several times so we don't have to squash, I can do so again when we figure this out.

I'll poke at this some more in a bit.

alerque:
Two things aren't clear to me:

* One it isn't clear that Reviewdog isn't broken. I tested elsewhere and it works for PR's originating from the same repo, but I couldn't get a test case across forks to work even with my own secrets. I'll look into that upstream.
* Also I can't tell if Actions isn't working as documented, if they changed something recently since I last set one of these up, or if Richard possibly cleared the secrets from the repo since last night just to be safe. I can't get any of the ones shown in the screenshot to give me data. I suspect that's because GitHub has further locked down the data PRs can access, which would mean we just can't have nice things.

The upshot is that I've set this up without a special special secret at all in such a way that it will still check PRs but throw the warnings to the log as usual, and then *try* to post PR reviews but not fail if it can't. That should give nicer things to first party PRs but work at some level for everybody.

Better things to do with the rest of my day, lets see how this works on the next fresh PR victim.

alerque (approved):

RichiH:
The tokens are still active as I didn't want to block any testing; if you want, I can delete them now or let them active for a week or so for testing; either works.

alerque:
Okay good to know. Maybe leave them for a bit, I'll poke in my fork and test with a new PR if I come up with anything.

RichiH:
GH support confirmed that it's a UI/UX issue and that I will not be oversharing access to reviewdog when giving access to my account; it's still limited to this repo. I updated as such:
image

Issue #294, opened 2021-04-05
`master` is hardcoded as default remote branch


RichiH:
This is surprisingly intricate.

https://github.com/RichiH/vcsh/blob/776d3f910aefe26e621b26ccb46474621e5b32bc/vcsh#L620 is an easy change assuming `main` as the local.

https://github.com/RichiH/vcsh/blob/776d3f910aefe26e621b26ccb46474621e5b32bc/vcsh#L176 might need a check for `main`, then `trunk`, then `master`, because of Git's inability to signal what the canonical remote branch is.

RichiH:
In the spirit of making things even more complicated, Git also suggests `development` because of course it did.

````
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
````

alerque:
I think there is a way to get the default remote branch name, but maybe only while you are cloning...

RichiH:
`git symbolic-ref refs/remotes/origin/HEAD` seems to work, but only after a remote has been set.

PR #295, opened 2021-04-05
Make remote branch finding more robust

RichiH:
Fixes: https://github.com/RichiH/vcsh/issues/294

This patch does three things at once:

1. Default to trying `main` first
2. Walk through several remote branches, in order
3. Test remote branches a lot earlier than before, aborting earlier

Signed-off-by: Richard Hartmann <richih@richih.org>


alerque (changes requested):
I don't think this is the best approach at all, we shouldn't care at all what branches are setup and iterating over a hard coded list is worse than the previous fallback to one hard coded option. In my case it will fail more often that the current because I have `development` branches that are **not my default branch**.

Why do we need to know this at all? For `init()` the user's Git config should take care of setting a default. For `clone()` doing a clone already does the right thing and pulls the default remote branch even if it's named *fooBarBazBob*. The only place we were using a branch name too early is in checking if the repository is empty, and we can do that with `git ls-remote "$VCSH_REMOTE" HEAD` instead of a branch name. We don't even need to parse the answer, just knowing there is a HEAD is enough and we can go ahead with the close using no branch name at all.

RichiH:
The current code does not work with `vcsh clone` and remote `main` branch which is what GitHub uses as default for new repos; ran into that migrating some repos with broken remotes over to GH.

I went off the list of branch names Git suggests; we could drop `development` as it's kinda stupid as the _default_ anyway.

But something I didn't consider: `git ls-remote "$VCSH_REMOTE" HEAD` and then simply `git clone`. I know that didn't work a decade ago which is why `clone()` is building a clone step-by-step, but maybe that changed. Assuming that works: I would be fine dropping the old code completely or moving it into a `clone_legacy()` in case someone has an ancient system they want to play with.

alerque:
I've never heard of (or at least don't remember) any Git server that couldn't handle `git clone <address>` without any branch information. I suspect the issue here is not that `git clone` won't work that way, it's that we aren't using `git clone` at all: the current clone function does our own init (including hooks), then uses `git fetch`. The right thing to do here is probably to figure out how to decouple that. I wouldn't realy expect `pre/post-init` hooks to run on `vcsh clone` for example.

RichiH:
The problem was (is?) that `WORKTREE` and such were not honored during clone; which forced me to clone by hand. As vcsh kicked off a bit of a buzz in Git circles back then, a lot of work was done around WORKTREE shortly after. I will try if setting WORKTREE before cloning does the right thing, these days.

RichiH:
More testing, trying to figure out what forced me to do it by hand back then:

````
Initialized empty Git repository in /home/richih/.config/vcsh/repo.d/git.git/
vcsh: info: remote branch 'main' found
Switched to a new branch 'main'
remote: Enumerating objects: 24, done.
remote: Counting objects: 100% (24/24), done.
remote: Compressing objects: 100% (15/15), done.
Unpacking objects: 100% (24/24), 2.57 KiB | 262.00 KiB/s, done.
remote: Total 24 (delta 3), reused 24 (delta 3), pack-reused 0
From github.com:RichiH/gitrc
 * branch            main       -> FETCH_HEAD
 * [new branch]      main       -> origin/main
error: Untracked working tree file '.gitconfig' would be overwritten by merge.
vcsh: fatal: will stop after fetching and not try to merge!
  Once this situation has been resolved, run 'vcsh git pull' to finish cloning.
````

We can't assume that the local `$HOME` is clean from cloned files and need to give useful warnings/errors as `git` is not built to anticipate this.

alerque:
Wouldn't that be better as bare clone first, then checkout the working tree later?

RichiH:
I can certainly try; spent today on unbreaking Debian packaging, and polishing up metamonger. Might take me until the weekend to get time to test it, but will do.

alerque:
Given that we want to see this make the v2 cut, I just handled the merge conflicts and added another tidbit I missed before: that `wc` is a dependency. Otherwise I haven't addressed the issues raised in comments here yet but I'll try to look into it too.

alerque (changes requested):
In reviewing the branch handling (and deciding to maybe not set `$VCSH_BRANCH` at all if the user didn't request it, I notices that L341 is using this variable inside `push()` and `pull()` subcommands, but the only place we're even setting it is in `clone()` subcommands. Something is wrong here.

RichiH:
It's coming in through https://github.com/RichiH/vcsh/pull/253 and it breaks both push and pull. Sorry, I should have caught this during PR review.

https://github.com/RichiH/vcsh/pull/253 to fix.

alerque (approved):
With #302 taking care of the other problem the rest of this looks good. Not perfect because it doesn't actually do remote default branch discovery, but since I can't figure out a way to do that yet this is definitely better than not having any remote branch latitude at all.

PR #296, opened 2021-04-11
Preload git completions

jamesd:
Two commits:
* Make sure git completions are loaded
* Add EditorConfig for _vcsh_bash

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,6 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 
-[vcsh]
+[{vcsh,_vcsh_bash}]
 indent_style = tab
 trim_trailing_whitespace = true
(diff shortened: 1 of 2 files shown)


alerque (approved):
Thanks for the contribution. I'll assume this works as intended on `bash`, I'm on `zsh` and I'm not sure my Bash installation isn't foobared with the recent experiments I was doing.

The Editor Config stuff is duplicated by #288 which both shuffled files around and adds this config, but I guess I can sort out the merge conflict there when it happens, it's perfectly reasonable to include it here while there is activity editing the file.

alerque (changes requested):
The only caveat is if you do include the Editor Config setup it does need to pass the lint tests. In my PR I resolve this by just joining that offending line into one long line instead of splitting it and having to worry about indentation. You could either resolve this the same way or drop the editor config commit, I don't care which way it goes.

alerque (approved):

PR #298, opened 2021-05-29
Collect odds and ends from forks that never got merged

alerque:
Original sources or inspiration:

* 9357db5 on @Dieterbe's fork.

### DIFF
--- a/vcsh.in
+++ b/vcsh.in
@@ -575,8 +575,7 @@ write_gitignore() {
 	done
 	IFS=$OLDIFS
 	if diff -N "$tempfile" "$VCSH_BASE/.gitignore.d/$VCSH_REPO_NAME" > /dev/null; then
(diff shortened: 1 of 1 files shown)


alerque:
I just combed through commits on a couple dozen forks and of the useful stuff I saw, all of it has been integrated in some way or another. Most of it was different people fixing the same problems multiple ways. Some seem to have triggered other work that did get merged. In any case this was the only tidbit I found (that isn't also in another PR for review). There could be more, but the network graph is crazy had to navigate.

alerque:
Oops! It turns out this one "tidbit" also happened to be wrong and the cause of #310.

> * remove a pointless `return 0` call

The removed return call wasn't pointless. The rest of the function logic *should* be skipped if there was nothing to do, and will fail since we've just cleaned up the tempfile early.

PR #299, opened 2021-06-16
Build standalone script edition in CI

### LINKED ISSUE #202: One-liner bootstrap

alerque:
Fixes #202.

I debated whether to call this special build "bootstrap" or "standalone". In the end "standalone" made more sense to me, but if anybody has a stronger opinion mine is pretty weak.

The idea is to just use the existing build system but tweak it to:

* not build any of the extras like man pages or completions
* not actually detect anything about the host environment, have it just assume `$PATH` will work at runtime
* add a `-standalone` segment to the version string so bug reports are easier to identify
* post the resulting script as an artifact on all CI jobs
* attach the resulting script "bare" (with no version info in the filename) to releases so that it can be downloaded using a oneliner, e.g. `curl -fsLOJS https://github.com/RichiH/vcsh/releases/latest/download/vcsh-standalone.sh`

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -2,6 +2,7 @@
 
 Many distributions have packages ready to go.
 If yours doesn't, you can install [from source](#installing-from-source).
+VCSH can also be deployed as a [standalone script](#standalone-script).
 If you package VCSH for a distro please let us know.
 
 ## Arch Linux
@@ -123,5 +124,30 @@ $ make DESTDIR="$HOME" install-exec
 ```
 
 This will install to `~/bin/vcsh`; add `~/bin` to your path to use.
+
+# Standalone Script
+
+A special variant of VCSH can be deployed as a single POSIX script with no configure/build step.
+Deploying it this way leave you without any man page or shell completion functions.
+This variant is also dependent or you `$PATH` to have proper versions of dependencies such as `git`.
+If your user space has different tools by default than your system beware!
+
+The standalone variant can be downloaded from under any entry in [releases](https://github.com/RichiH/vcsh/releases).
+
+This method is suited for installation to a user space where you don't have control over the system packages, e.g.:
+
+```console
+$ mkdir ~/bin
+$ curl -fsLS https://github.com/RichiH/vcsh/releases/latest/download/vcsh-standalone.sh -O ~/bin/vcsh
+$ chmod 755 ~/bin/vcsh
+```
+
+It could also be used to directly bootstrap a dotfiles repository something like this:
+
+```console
+$ sh <(curl -fsLS https://github.com/RichiH/vcsh/releases/latest/download/vcsh-standalone.sh) clone <path_to_your_dotfiles_repo> dotfiles
+```
+
+Note that we strongly encourage using a tagged version that you've tested to work for you instead of the *latest* keyword.
 
 [1]: http://rtomayko.github.io/ronn/
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -127,12 +127,12 @@ This will install to `~/bin/vcsh`; add `~/bin` to your path to use.
 
 # Standalone Script
 
-A special variant of VCSH can be deployed as a single POSIX script with no configure/build step.
-Deploying it this way leave you without any man page or shell completion functions.
-This variant is also dependent or you `$PATH` to have proper versions of dependencies such as `git`.
+A special variant of VCSH can be deployed as a single POSIX shell script with no configure/build step.
+Deploying it this way leaves you without any man page or shell completion functions (or possibly with mismatched resources installed by your package manager for a different vcsh version).
+This variant is also dependent or your `$PATH` to have proper versions of dependencies such as `git`.
 If your user space has different tools by default than your system beware!
 
-The standalone variant can be downloaded from under any entry in [releases](https://github.com/RichiH/vcsh/releases).
+The standalone variant can be downloaded from any recent entry in [releases](https://github.com/RichiH/vcsh/releases).
 
 This method is suited for installation to a user space where you don't have control over the system packages, e.g.:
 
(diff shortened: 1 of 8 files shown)


alerque:
Unfortunately due to the way GitHub Actions is setup the standalone version artifact attached to each CI run here can only be downloaded as a zip file. You can see it as built for this PR [here](https://github.com/RichiH/vcsh/actions/runs/942544643). This isn't very conducive to oneliners, but I think that's okay. If people are going to script this they shouldn't be using some random CI run artifact anyway.

The standalone script attached to each release should download as a bare script ready for execution by a shell. It's kind of hard to test that until we actually tag a release, but I'll try to do a prerelease on my fork and see if it works there.

RichiH:
What about "hosting" on https://pages.github.com/ ? We could have a shell script there.

Minor nit, `@DEPLOYMENT@` might make more sense, as that could also cover e.g. `vcsh-git` for development.

alerque:
Yes, this could be hosted on GitHub pages. What would be the advantage to that over attaching the script to releases? The disadvantage I see in moving to Pages is the disconnect from the tagged versions. Of course we could host multiple versions there with something matching the tag versions, but with artifacts attached to releases the relation to the release cycle seems more clear and it would encourage people to use a tagged version while still giving the option of using the `latest` keyword in place of a tag version.

alerque:
> Minor nit, `@DEPLOYMENT@` might make more sense, as that could also cover e.g. `vcsh-git` for development.

I don't have a problem with `@DEPLOYMENT@` over `@EDITION@`, but it would have no effect on development versioning. Those already have an elaborate version string with the last tag, the number of commits since the tag, and the commit hash. We don't need anything else to identify them, this would be purely for identifying the type of installation that was used.

RichiH:
> Yes, this could be hosted on GitHub pages. What would be the advantage to that over attaching the script to releases

That it's `vcsh-standalone.sh`, not `vcsh-standalone.zip`. It should always point to latest, agreed.

RichiH:
> does ./configure --with-deployment=standalone make more sense than ./configure --with-edition=standalone?

I think it makes _slightly_ more sense, yes.

alerque:
> > Yes, this could be hosted on GitHub pages. What would be the advantage to that over attaching the script to releases
> 
> That it's `vcsh-standalone.sh`, not `vcsh-standalone.zip`. It should always point to latest, agreed.

The standalone scripts attachments to releases will not be ZIPs anyway, they should download as plain text ready for reading by a POSIX shell. That caveat only applied to the CI runs for every commit/PR.

alerque:
For a real preview of what I expect this to look like, I tagged a sample release on my fork. Check out [this simulated v1.9.0 release](https://github.com/alerque/vcsh/releases/tag/v1.9.0). See the artifacts? Is that clear that there is a source release (tar.xz and zip) plus a standalone raw shell script attached? For testing you should be able to `curl -fsLJS https://github.com/alerque/vcsh/releases/download/v1.9.0/vcsh-standalone.sh` for direct download of the script via the tag or `curl -fsLJS https://github.com/alerque/vcsh/releases/download/latest/vcsh-standalone.sh` for the *latest* keyword version with no tag name required.

alerque:
Sorry for the noisy pushes in this PR, most of them don't have any bearing on the discussion above. I was using this as as excuse to shake down the CI jobs handling lint/test/build/release chores. The process seems pretty solid now, for example here is a [simulated v1.9.1 release](https://github.com/alerque/vcsh/releases/tag/v1.9.1) where the build artifacts (source dist plus standalone script) as well as the changelog notes are all posted automatically.

doc/INSTALL.md
```diff
@@ -124,4 +125,29 @@ $ make DESTDIR="$HOME" install-exec
 
 This will install to `~/bin/vcsh`; add `~/bin` to your path to use.
 
+# Standalone Script
+
+A special variant of VCSH can be deployed as a single POSIX script with no configure/build step.
```
RichiH, line 130:
```suggestion
A special variant of VCSH can be deployed as a single POSIX shell script with no configure/build step.
```

doc/INSTALL.md
```diff
@@ -124,4 +125,29 @@ $ make DESTDIR="$HOME" install-exec
...
+Deploying it this way leave you without any man page or shell completion functions.
```
RichiH, line 131:
```suggestion
Deploying it this way leaves you without any man page or shell completion functions, or the versions installed by your package manager for a different vcsh version.
```

doc/INSTALL.md
```diff
@@ -124,4 +125,29 @@ $ make DESTDIR="$HOME" install-exec
...
+This variant is also dependent or you `$PATH` to have proper versions of dependencies such as `git`.
+If your user space has different tools by default than your system beware!
+
+The standalone variant can be downloaded from under any entry in [releases](https://github.com/RichiH/vcsh/releases).
```
RichiH, line 135:
```suggestion
The standalone variant can be downloaded from any recent entry in [releases](https://github.com/RichiH/vcsh/releases).
```

doc/INSTALL.md
```diff
@@ -124,4 +125,29 @@ $ make DESTDIR="$HOME" install-exec
...
+
+This method is suited for installation to a user space where you don't have control over the system packages, e.g.:
+
+```console
+$ mkdir ~/bin
+$ curl -fsLS https://github.com/RichiH/vcsh/releases/latest/download/vcsh-standalone.sh -O ~/bin/vcsh
+$ chmod 755 ~/bin/vcsh
+```
+
+It could also be used to directly bootstrap a dotfiles repository something like this:
```
RichiH, line 145:
```suggestion
It could also be used to directly bootstrap a dotfiles repository with something like this:
```

doc/INSTALL.md
```diff
@@ -124,4 +125,29 @@ $ make DESTDIR="$HOME" install-exec
...
+
+```console
+$ sh <(curl -fsLS https://github.com/RichiH/vcsh/releases/latest/download/vcsh-standalone.sh) clone <path_to_your_dotfiles_repo> dotfiles
+```
+
+Note that we strongly encourage using a tagged version that you've tested to work for you instead of the *latest* keyword.
```
RichiH, line 151:
```suggestion
While we are enabling cURL-based workflows on purpose, we still encourage you to avoid them where reasonably possible. If you do use it, please consider using a tagged version that you've tested to work for you instead of the "latest" keyword.
```

RichiH (approved):
Minor nits, else LGTM, thanks!

RichiH:
It's a tad noisy but my workflow hides most of this, no worries. And thanks for actually doing the work! :)

nferch:
Thanks @alerque and @RichiH for all of the work and review on this!

Sorry if I missed this, but is there an issue or PR to track the first release that will contain this and make https://github.com/RichiH/vcsh/blob/main/doc/INSTALL.md reflect reality?

RichiH:
Good point. https://github.com/RichiH/vcsh/issues/301 is the tracking issue which will result in a new release being built.

PR #326, opened 2022-10-18
Remove bashism from configure script

### LINKED ISSUE #325: configure fails with non-Bash /bin/sh
`./configure` when run with a non-Bash `/bin/sh` emits an error:
```
checking for x86_64-pc-linux-gnu-pkg-config... /usr/sbin/x86_64-pc-linux-gnu-pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for BASH_COMPLETION... yes
./configure: 3430: Bad substitution
checking that generated files are newer than configure... done
configure: creating ./config.status
[...]
```

This is because the definition of TRANSFORMED_PACKAGE_NAME at https://github.com/RichiH/vcsh/blob/main/configure.ac#L108 relies on a Bashism.

alerque:
Fixes #325

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Fetch tags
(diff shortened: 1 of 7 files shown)


alerque:
@thesamesam Does this do the trick?

Of course the usual `./configure` to get defaults should work, but it would be nice to know if the substitutions still work too outside of a bash environment. Something like this should check on that:

```console
$ ./configure --program-prefix=foo --program-suffix=bar --prefix /tmp/foo
$ make install
$ stat /tmp/foo/bin/foovcshbar
```

thesamesam:
Sorry, I'd missed this! I'll test it in about half an hour

thesamesam:
Looks great, thank you!

PR #334, opened 2023-12-28
Revert change to .gitignore.d behavior in #196

### LINKED ISSUE #333: .gitignore.d is broken for files in subdirs due to git behavior revert
**Repro steps**
1. Enter a vcsh-controlled repository, e.g. via `vcsh foo`
2. Ensure that `.gitignore.d/foo` exists and is up-to-date. Update via `vcsh write-gitignore foo` if needed.
3. `git add <some modified file, in a subdirectory and already tracked by foo and in the above gitignore>`

**Expected results**
`git add` just works.

**Actual results**
`git add` produces the "force me" error message, such as the following for a file under `~/.config`:
```
The following paths are ignored by one of your .gitignore files:
.config
```

Ugh, it looks like git [reverted the enhancement regarding gitignore parent directories](https://github.com/git/git/commit/5cee349370bd2dce48d0d653ab4ce99bb79a3415), and we need to resurrect the old code that added each parent directory of a controlled file into `.gitignore.d/foo`:

```
*
!/.config
!/.config/bar
!/.config/bar/baz
!/.config/bar/baz/something.txt
```

**Notes**
The 2.43.0 gitignore man page includes this note, which I used to trace back to the revert commit above:
> An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined. 

**Versions**
vcsh: 2.0.5 standalone
git: 2.43.0

jwhitley:
Fixes #333.

In git 2.7.0, new git behavior changed gitignore to no longer require un-ignoring every parent directory above an un-ignored file. vcsh updated its logic around this in PR #196 (RichiH/vcsh@e4f4ecf). Unfortunately, git's change was reverted before 2.8.0-rc4 (git/git@5cee349).

This change therefore reverts the logic change in #196 causing `vcsh write-gitignore` to once again emit parent directories into generated `.gitignore.d/<repo>` files.

See #333 for the gory details.

### DIFF
--- a/vcsh.in
+++ b/vcsh.in
@@ -569,16 +569,11 @@ write_gitignore() {
 	OLDIFS=$IFS
 	IFS=$(printf '\n\t')
 	gitignores=$(for file in $(@GIT@ ls-files); do
-		if [ "$GIT_VERSION_MAJOR" -ge 2 ] \
-				&& [ "$GIT_VERSION_MINOR" -ge 7 ]; then
-			echo "$file";
-		else
-			while true; do
-				echo "$file"; new=${file%/*}
-				[ x"$file" = x"$new" ] && break
-				file=$new
-			done;
-		fi
+		while true; do
+			echo "$file"; new=${file%/*}
+			[ x"$file" = x"$new" ] && break
+			file=$new
+		done;
 	done | sort -u)
 
 	# Contrary to GNU mktemp, mktemp on BSD/OSX requires a template for temp files


jwhitley:
FWIW, I opted to leave the code that gets git's version info, but that's now dead code so there's an argument to remove it as well.

jwhitley:
The editorconfig lint check appears to have failed because it's _looking inside the `.git` directory?!_. Looks like something's misconfigured.

jwhitley:
Confirmed. Looks like an update to the editorconfig checker Docker image is required, to >=2.7.1. Might as well bump to latest release of 2.7.2. See editorconfig-checker/editorconfig-checker#268

alerque (approved):

PR #342, opened 2024-03-05
Add suport for adding main commands, plugins

hansmansson:
vcsh is incredibly extendable and flexible however there is no possibility to add new commands. This is were plugins come into play. By adding plugins to $XDG_CONFIG_HOME/vcsh/plugins-enabled new commands become available.
The plugin file name becomes the new command, the file must contain a function with the same name as the file, this is the entry point. A help function can also be added to get append the plugin help text to the main help text, if help is omitted a default help message is added.

### DIFF
--- a/doc/vcsh.1.ronn.in
+++ b/doc/vcsh.1.ronn.in
@@ -300,6 +300,19 @@ may even lose data.
 
 You have been warned.
 
+## PLUGIN SYSTEM
+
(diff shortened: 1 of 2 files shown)


alerque:
Implementation wise this looks pretty good. Can you give an example of an actual use case? When do you need more access to internals while a command runs than `vcsh run` provides?

